PR #168, opened 2014-09-28
First attempt at socket-idle-timeout implementation.

NoamB:
Hi,
This PR is not ready for merge (no tests) but I wanted to show it to you in a readable way (diffs).

I've added my own data structure called ActivityCollection, that is built from a HashMap and a self-made double-linked list.
It allows constant time for all the actions needed for this feature (add, remove, update).

The ActivityCollection instance is not created if the user leaves the options at default, and so there is no overhead if you don't need the feature. To make the code nicer, instead adding an "if" everywhere, I've added a NoopActivityCollection which shares an interface with ActivityCollection and does nothing in all calls to it.

If you do use the feature it costs a few more references and one long per key, RAM-wise.

All the existing tests are still passing.

How it works - 
On accept - keys are inserted into the ActivityCollection, at the head, insertion timestamp saved.
On read/write - keys are updated, effectively "bumped" to the head of the list, and their timestamp is updated.
On run loop - keys are taken from the tail of ActivityCollection in a loop, and we call closeKey() on them until we find one that isn't expired or we run out of keys. This means we expire keys in O(m), m being the number of keys that have expired.

It needs no synchronisation as it is only called from the server thread.


ptaoussanis:
Closing this because of inactivity. Please feel free to reopen as a new PR against the current master if you'd still like to pursue this - thanks!

PR #178, opened 2014-12-06
Fixing fragmented WebSocket messages

bitnovus:
Frames with a payload length of 127 would cause WSDecoder to throw an
exception after going from PAYLOAD to FRAME_START as the opcode would
not match the conditions in FRAME_START. The fix was to set the variable 
"opcode" to -1 before returning to state FRAME_START. Another exception 
would be thrown on reading the final frame due to the first frame's 
opcode not being preserved. The fix was to add the variable 
"fragmentedOpCode" that would preserve the first frame's opcode. The 
changes should be in accordance with Section 5.4 of RFC 6455.

The ProtocolException in READ_8_LENGTH had a misleading statement. For
example, the max payload length can be greater than 4M. It now provides
a more informative message.

### DIFF
--- a/src/java/org/httpkit/server/WSDecoder.java
+++ b/src/java/org/httpkit/server/WSDecoder.java
@@ -29,6 +29,7 @@ public class WSDecoder {
     private int maskingKey;
     private boolean finalFlag;
     private int opcode = -1;
+    private int fragmentedOpCode = -1;
     private int framePayloadIndex; // masking per frame
 
     // 8 bytes are enough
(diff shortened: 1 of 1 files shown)


ldenman:
@weltan @bitnovus I am so glad this was fixed as well. Thanks! I can confirm this resolves the trouble I had using weasel and sending ~ 1MB payloads. :+1:

olivergeorge:
Yay

pandeiro:
when will this fix make it in to a release?

ptaoussanis:
Should be available now as [com.taoensso.forks/http-kit "2.1.20"] (https://github.com/http-kit/http-kit/releases)

Issue #90, opened 2013-11-03
http-kit buffers whole response body from InputStream before sending to client


eduardoejp:
If I'm using a PipedOutputStream to send data to a corresponding PipedInputStream, hoping for the client to download a large file that I'll be sending to it in small pieces, http-kit doesn't send anything until all the data has been written to the stream.

That creates 2 problems:
1. If I send big files many times simultaneously, I can run out of memory.
2. Having to wait until all data has been written to the stream introduces latency for big files.

flyingmachine:
I wrote [some code](https://github.com/flyingmachine/streaming-proxy) that should help with writing a smart proxy :smile: It uses clj-http to stream responses.

bhurlow:
@ptaoussanis @shenfeng Hi all! What's the status on this issue. Still not possible for the http-kit client to consume response chunks?

ptaoussanis:
Hi Brian, the http-kit issues are currently up-to-date so if an issue's still open means it's unfortunately still unresolved.

Most larger bug fixes are currently blocked on http-kit's author (@shenfeng) who's been unavailable. Hoping that he'll find some time in future to return to http-kit for the bug fixes.

If not, then I (or someone else) will need to try find time to properly familiarize with the code base. In the meantime, I'm accepting simple PRs.

Hope that helps!

bhurlow:
@ptaoussanis thnx!

dgrnbrg:
Fwiw, we ended up using jet in the project that needed this functionality.

bhurlow:
@dgrnbrg I ended up using aleph. can you link me to jet?

dgrnbrg:
https://github.com/mpenet/jet We may have had to tweak it a bit, but we managed to get streaming, chunking client & server side, and we say 7-8G throughput on a 10G ethernet link. So it worked very well for high performance streaming.

bhurlow:
Awesome!! @dgrnbrg thnx

ptaoussanis:
## Update: Mar 2016

Leaving this issue (along with #202) open for now but for anyone who might come across this please note that these are **unlikely to be resolved in the near future w/o a specific PR**.

To [quote](https://github.com/http-kit/http-kit/issues/90#issuecomment-29497954) @shenfeng:
"http-kit is not good at large file download(since it buffer whole response in memory). clj-http is a good alternative and http-kit's client API is modeled after clj-http's. clj-http's response can be a stream."

sundbry:
[...]

PR #181, opened 2014-12-16
Call handler without waiting for entire body

martinraison:
We were facing issue #90 where our server would wait for the entire body to be received before calling the handler. Since the body is passed to the handler as an `InputStream`, it makes more sense to have the handler being called right after the headers are decoded.

Here is a fix along with a new test case. We had to modify the handler for `test-chunked-encoding`, because in the case of a chunked request, the content-length is still unknown by the time the handler is called.


ptaoussanis:
Closing this because of inactivity. Please feel free to reopen as a new PR against the current master if anyone would still like to pursue this - thanks!

martinraison:
All right, I don't use http-kit anymore so I'm unlikely to update this PR.

PR #185, opened 2015-01-19
Do not set client mode on SSLEngine if already set.

izarov:
Prevent exception from being thrown calling setUseClientMode on a user-supplied client mode SSLEngine.


ptaoussanis:
Thank you!

PR #190, opened 2015-01-29
Transfer-Encoding is only supported in HTTP Version 1.1.

jenshaase:
Therefore the transfer-encoding header is only send in this version.

### DIFF
--- a/src/java/org/httpkit/server/AsyncChannel.java
+++ b/src/java/org/httpkit/server/AsyncChannel.java
@@ -114,7 +114,9 @@ public class AsyncChannel {
         if (close) { // normal response, Content-Length. Every http client understand it
             buffers = HttpEncode(status, headers, body);
         } else {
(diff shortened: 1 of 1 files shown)


ptaoussanis:
Hi Jens, thank you!

Has this change been tested in production? Do you believe it should be safe to merge? No possible negative effects that you're aware of?

mhjort:
At least all the tests pass with this change and this is how it should work by the spec. Current implementation is incorrect in that sense. I think it would be pretty safe to merge this in.

venantius:
Agree. Merging.

ptaoussanis:
My concern here on a skim is that the PR appears to affect only the header; it's not clear to me that this is having the desired effect on the actual encoding?

Has someone confirmed that with the header removed, http-kit actually provides non-chunked encoding?

ptaoussanis:
Just confirmed that `HttpEncode` is indeed checking for the chunked header.

PR #201, opened 2015-04-08
client: option to keep "unsafe" method on redirect

dzaharee:
Currently the client changes the request method to GET when following a
301, 302, or 303 redirect. This adds an :allow-unsafe-redirect-methods
request option that can be used to override this behaviour.


ptaoussanis:
Thanks David!

PR #211, opened 2015-06-06
add flash-policy-file request responder

zoowii:
I found as if the library does not support the passage of flash-policy-file to supply WebSocket browser does not support flash to a compatible WebSocket through the API, so I modify the code to support flash-policy-file responder. user can add :allow-access-uris ["http://example.com:843"] in run-server params

### DIFF
--- a/src/java/org/httpkit/FlashPolicyFileRequestFoundExceiption.java
+++ b/src/java/org/httpkit/FlashPolicyFileRequestFoundExceiption.java
@@ -1,0 +1,10 @@
+package org.httpkit;
+
+public class FlashPolicyFileRequestFoundExceiption extends HTTPException {
+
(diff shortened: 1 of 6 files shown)


shenfeng:
Hello, you can do that in a middleware? like  `wrap-keyword-params`

zoowii:
When use flash to support IE8 WebSocket, the server need response a special socket request once sent from flash, This request is not Http and not WebSocket, so I can use a proxy server like nginx before http-kit, or directly change the http-kit request-decoder the support. I don't know how to use a middleware to process socket/channel directly before http-kit handler. Do you have any example?

ptaoussanis:
Closing this because of inactivity. Please feel free to reopen as a new PR against the current master if anyone would still like to pursue this - thanks!

PR #216, opened 2015-06-19
Remove unnecessary operation.

songsd:
when iterating the set,it’s not necessary to produce a array and then
iterate.


ptaoussanis:
Thanks!

PR #217, opened 2015-06-19
More accurate timeout handling

mbarbon:
With the current implementation, the actual connect timeout can be up to 2 seconds greater than the passed-in connection timeout; the patch tries to address this by using the time remaining for the the first queued request to expire as the actual timeout.

### DIFF
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -361,8 +361,13 @@ public final class HttpClient implements Runnable {
     public void run() {
         while (running) {
             try {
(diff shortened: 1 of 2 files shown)


ptaoussanis:
Hi Mattia, thank you!

Has this change been tested in production? Do you believe it should be safe to merge?

mbarbon:
It worked for my test case; since then we switched away from http-kit, so I don't know if it would cause other issues on the longer run/with a different workload.

ptaoussanis:
Thanks, will take a look at including this in the next release :-)

PR #223, opened 2015-07-11
updated to display latest Clojars version

yogthos:


yogthos:
just a heads up, the home page has the old version as well

PR #225, opened 2015-07-21
LinkedRunnable.run: don't consume stack unnecessarily

Chouser:
Sending lots of small messages at a high rate to an http-kit websocket will sometimes create a chain of LinkedRunnable's long enough to cause a stack overflow on this line:

https://github.com/http-kit/http-kit/blob/73f82165c73d2a0599e610db8cb7a50de48d31cd/src/java/org/httpkit/server/RingHandler.java#L122

This pull request implements a loop to follow the chain in a single stack frame, avoiding the stack overflow.

### DIFF
--- a/src/java/org/httpkit/server/RingHandler.java
+++ b/src/java/org/httpkit/server/RingHandler.java
@@ -118,8 +118,12 @@ class LinkingRunnable implements Runnable {
 
     public void run() {
         impl.run();
-        if (!next.compareAndSet(null, this)) { // has more job to run
-            next.get().run();
+
+        // Run all jobs in this chain without consuming extra call stack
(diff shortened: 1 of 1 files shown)


shenfeng:
`next.get().run()` will run the reset,  but  `r.impl.run()` will not. 



a -> b -> c

we are at `a`,  ``next.get().run()`` will run  `b`, then `c`,  `r.impl.run()` will only run `b`.

Chouser:
This is why `r.impl.run()` is called in a loop, so that the run of `b` and `c` will be called, and will be called in the right order.

shenfeng:
recursion -> loop

Thanks.

Chouser:
Exactly. Thanks!

Issue #194, opened 2015-02-26
Hangs when using http streaming channels under load of requests


michaelr524:
Maybe related this issue: https://github.com/http-kit/http-kit/issues/76

Even when the handler doesn't do anything but just closes the channel, it looks like all threads are waiting for something.

Tried with latest code from github master branch and also with 2.1.16 and 2.1.19 from clojars.
Is anyone familiar with this problem?

ptaoussanis:
Hi there, busy doing triage - is this still an issue? Is it still reproducible with the [latest release](https://github.com/http-kit/http-kit/releases)?

michaelr524:
Hi Peter, I've replaced httpkit with immuntant ever since so can't really tell..

ptaoussanis:
Great, thanks for the confirmation :-)

Issue #118, opened 2014-02-11
http-kit client can get in a stuck state (blocking) after heavy use


violahs:
I have written a multithread news aggregator in Clojure.  http-kit has been awesome for this...  so far.  Recently,  I have been able to get http-kit 2.1.16 into a stuck state (perhaps it is out of threads??).

After a while my normal pipeline begins to block.   The requests never return.

I can then pop over to the REPL and submit the code below which blocks forever.  I eventually ^c out of it, and it then complains about thread death.  (I am on the net.  I can connect using the same REPL using clj-http.)

```
eg-feedparser.feeds> (def rr @(http/get "http://www.google.com"))
CompilerException java.lang.ThreadDeath, compiling:(form-init8115144213294051224.clj:1:1) 
eg-feedparser.feeds> 
```

Below is more of the stack trace.

```
clojure.lang.Compiler$CompilerException: java.lang.ThreadDeath, compiling:(form-init8115144213294051224.clj:1:1)
 at clojure.lang.Compiler$InvokeExpr.eval (Compiler.java:3463)
    clojure.lang.Compiler$DefExpr.eval (Compiler.java:408)
    clojure.lang.Compiler.eval (Compiler.java:6624)
    clojure.lang.Compiler.eval (Compiler.java:6582)
    clojure.core$eval.invoke (core.clj:2852)
    clojure.main$repl$read_eval_print__6588$fn__6591.invoke (main.clj:259)
    clojure.main$repl$read_eval_print__6588.invoke (main.clj:259)
    clojure.main$repl$fn__6597.invoke (main.clj:277)
    clojure.main$repl.doInvoke (main.clj:277)
    clojure.lang.RestFn.invoke (RestFn.java:1096)
    clojure.tools.nrepl.middleware.interruptible_eval$evaluate$fn__872.invoke (interruptible_eval.clj:56)
    clojure.lang.AFn.applyToHelper (AFn.java:159)
    clojure.lang.AFn.applyTo (AFn.java:151)
    clojure.core$apply.invoke (core.clj:617)
    clojure.core$with_bindings_STAR_.doInvoke (core.clj:1788)
    clojure.lang.RestFn.invoke (RestFn.java:425)
    clojure.tools.nrepl.middleware.interruptible_eval$evaluate.invoke (interruptible_eval.clj:41)
    clojure.tools.nrepl.middleware.interruptible_eval$interruptible_eval$fn__913$fn__916.invoke (interruptible_eval.clj:171)
[...]

Issue #73, opened 2013-07-23
Exceptions when server is under load


jakepearson:
Hi,
I am seeing this exception periodically when the server is under load.  I believe it might happen when the handler is slow (maybe a queueing problem?).

    Tue Jul 23 11:42:37 MDT 2013 [server-loop] ERROR - http server loop error, should not happen
    java.nio.channels.CancelledKeyException
    	at sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:55)
    	at sun.nio.ch.SelectionKeyImpl.readyOps(SelectionKeyImpl.java:69)
    	at sun.nio.ch.KQueueSelectorImpl.updateSelectedKeys(KQueueSelectorImpl.java:105)
    	at sun.nio.ch.KQueueSelectorImpl.doSelect(KQueueSelectorImpl.java:74)
    	at sun.nio.ch.SelectorImpl.lockAndDoSelect(SelectorImpl.java:69)   
    	at sun.nio.ch.SelectorImpl.select(SelectorImpl.java:80)
    	at sun.nio.ch.SelectorImpl.select(SelectorImpl.java:84)
    	at org.httpkit.server.HttpServer.run(HttpServer.java:232)
    	at java.lang.Thread.run(Thread.java:680)

ptaoussanis:
Closing for now, please feel free to reopen if anyone's still having trouble with this using the [latest http-kit release](https://github.com/http-kit/http-kit/releases).

PR #234, opened 2015-10-22
Catch Throwable when delivering response

msassak:
The `deliver-resp` helper function in `org.httpkit.client/request` catches Exception, but not Throwable. Any code in the callback that throws Throwable rather than Exception prevents the client thread from ever returning (so far as I can tell). That is unfortunate, because Clojure's `assert`, and `:pre` and `:post` conditions all throw Throwable rather than Exception, leading to some very hard to troubleshoot failures.

I believe this might be the problem behind these issues:

* https://github.com/http-kit/http-kit/issues/73
* https://github.com/http-kit/http-kit/issues/118
* https://github.com/http-kit/http-kit/issues/194

If not, they at least appear to describe behavior in http-kit similar to what made me look into this.

### DIFF
--- a/src/org/httpkit/client.clj
+++ b/src/org/httpkit/client.clj
@@ -149,7 +149,7 @@
   (let [{:keys [url method headers body sslengine]} (coerce-req opts)
         deliver-resp #(deliver response ;; deliver the result
                                (try ((or callback identity) %1)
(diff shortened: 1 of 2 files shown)


ptaoussanis:
Thank you!

PR #236, opened 2015-10-22
Add Proxy Protocol support

tendant:
Add proxy protocol support. 

Some background about Proxy Protocol:

http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt

Since AWS ELB does not support WebSocket using HTTP protocol. The work around is to use TCP with PROXY Protocol. 

This patch will add PROXY Protocol support.


tendant:
@ptaoussanis The conflict has been resolved.

ptaoussanis:
Looks good Lei, thank you!

PR #239, opened 2015-11-04
HTTP client: keep form params in multipart requests

skazhy:
Currently, if a multipart request has both form params and files, the former are lost when coercing the request. This PR fixes the issue.


ptaoussanis:
Thanks for this!

PR #245, opened 2015-12-13
Replace run-server varargs with multiple arities

Deraen:
Varargs allows any number of arguments but in this case only one or two
arguments are valid.

http://stuartsierra.com/2015/06/01/clojure-donts-optional-arguments-with-varargs

### DIFF
--- a/src/org/httpkit/server.clj
+++ b/src/org/httpkit/server.clj
@@ -11,28 +11,28 @@
   param to wait existing requests to be finished, like (f :timeout 100).
 
   * See http://http-kit.org/migration.html for differences."
-  [handler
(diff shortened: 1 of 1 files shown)


ptaoussanis:
Hi Juho,

Thanks for this! Going to decline though, don't particularly agree that this is a problem. The `([] (self opts)) ([opts]))` form has benefits (performance + compiler arity checking), but it's also longer and more error prone in my experience. 

`& [opts]` as used here is a common idiom and one that I feel makes sense in cases that aren't easily confused and that aren't hurt by the destructuring cost. 

Appreciate the PR though, feel free to keep them coming! :+1: :-)

PR #248, opened 2016-01-22
Correctly reset fragmentedOpCode on WS frame completion.

daviesian:
When sending frames of different types (opcodes) over websockets to the httpkit server, the type of frames is incorrectly set after fragmentation occurs. This is because `fragmentedOpCode` is never reset in `WSDecoder.java`. This pull request fixes the bug.

To reproduce the problem:

* Send a text frame to the server
* Send a fragmented binary frame to the server
* Send a text frame to the server

Notice that the final text frame is delivered as a binary frame. This obviously shouldn't happen.

I would have liked to add a test to cover this case, but I'm afraid I couldn't work out how to force the `org.httpkit.ws.WebSocketClient` to fragment a binary frame in the necessary way. I'd be happy to add a test if someone could point me in the right direction.

### DIFF
--- a/src/java/org/httpkit/server/WSDecoder.java
+++ b/src/java/org/httpkit/server/WSDecoder.java
@@ -149,6 +149,7 @@ public class WSDecoder {
                         if (finalFlag) {
                             if (fragmentedOpCode > 0)
                               opcode = fragmentedOpCode;
+                            fragmentedOpCode = -1;
                             switch (opcode) {
                                 case OPCODE_TEXT:
                                     return new Frame.TextFrame(content);


ptaoussanis:
Hi Ian, thank you for this!

Would you consider this fix stable? Have you had this running in production?

daviesian:
I'm not sure I'd go so far as claiming it is stable, given that I haven't been able to write a test for it, but it is definitely better than it was. I am using it in production and have seen no problems so far!

ptaoussanis:
Okay, thanks- will include this in the next release I cut.

Please update here if you do happen to run into any issues at some point.

Cheers :-)

PR #254, opened 2016-02-03
Update HttpServer.java

Thingographist:
64Kb

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,17 +4,18 @@
 
 See **[http-kit.org](http://http-kit.org)** for documentation, examples, benchmarks, etc.
 
-Current [semantic](http://semver.org/) version:
+Current [semantic](http://semver.org/) versions:
+
 ```clojure
-[http-kit "2.1.19"] ; Latest official version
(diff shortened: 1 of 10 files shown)


ptaoussanis:
Hi there, could you explain a little what the motivation is here? Thanks

ptaoussanis:
Closing since this seems to be a duplicate of #236 (already merged)

ptaoussanis:
Merged, thanks!

hiteshjasani:
@Thingographist, @ptaoussanis  Sorry, I realize this has already been merged in, but can you comment on why this change was desired?

[RFC 1323](http://www.ietf.org/rfc/rfc1323.txt) defines Window Scale Options which

> The scale factor is limited to a power of two and encoded logarithmically,
> so it may be implemented by binary shift operations.

In other words, all major OS's provide the ability to define TCP Window Sizes beyond the original 16 bits originally provided in the header.  So TCP windows can indeed be truly `65,536 (2**16)` in size very easily.  Defining a capacity for this buffer that is not a power of two would seem to lead to additional buffer reads in some cases, a wasted byte to memory alignment padding and possible [misaligned memory access](https://www.kernel.org/doc/Documentation/unaligned-memory-access.txt).

Issue #247, opened 2016-01-22
Empty sequence response results in NullPointerException


leblowl:
Test subject:

```clojure
(defn test []
  '())

# compojure route
(GET "/test" [] (test))
```

Exception:
```shell
Thu Jan 21 20:12:31 PST 2016 [worker-1] ERROR - GET /test
java.lang.NullPointerException
	at org.httpkit.HttpUtils.bodyBuffer(HttpUtils.java:102)
	at org.httpkit.HttpUtils.HttpEncode(HttpUtils.java:425)
	at org.httpkit.server.HttpHandler.run(RingHandler.java:101)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```
I think it makes sense looking at the code in HttpUtils https://github.com/http-kit/http-kit/blob/b495e5fb3c44ac792654bcf4157993613bc579cd/src/java/org/httpkit/HttpUtils.java#L105-L115... feel free to let me know whatchya think

ptaoussanis:
Hi Lucas, thanks for the report. PR to fix would be welcome!

leblowl:
np, I will try to work on this, ty

PR #255, opened 2016-02-06
Fix NullPointerException on empty sequence response

leblowl:
Hola,

Here is my attempt @ fixing http-kit/http-kit#247. I also added a test route to the test server for this example. I am not sure that was the best place to put it, if there is a unit testing section for this, then I can add something there. p.s. It was a little weird transitioning from Clojure to Java... count in clojure doesn't throw null pointers ! Also I have a question, where does this count method even come from, the body gets cast to an ISeq and that interface doesn't contain any count method from what I can tell. Many thanks

### DIFF
--- a/test/org/httpkit/server_test.clj
+++ b/test/org/httpkit/server_test.clj
@@ -110,7 +110,10 @@
                               :body "Hello World"}))
   (GET "/iseq" [] (fn [req] {:status 200
                             :headers {"Content-Type" "text/plain"}
-                            :body (range 1 10)}))
+                             :body (range 1 10)}))
+  (GET "/iseq-empty" [] (fn [req] {:status 200
+                                   :headers {"Content-Type" "text/plain"}
+                                   :body '()}))
   (GET "/file" [] (wrap-file-info file-handler))
   (GET "/ws" [] (fn [req]
                   (with-channel req con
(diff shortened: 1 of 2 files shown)


leblowl:
Not sure what is up with the merge commit, if you have advice on how to prettify the history, I can do it...

ptaoussanis:
No problem re: history, I'll sort that out.

Am wondering though: is there a particular reason we can't just return null for an empty seq here instead of still going through the 0-count DynamicBytes?

test/org/httpkit/server_test.clj
```diff
@@ -110,7 +110,10 @@
                               :body "Hello World"}))
   (GET "/iseq" [] (fn [req] {:status 200
                             :headers {"Content-Type" "text/plain"}
```
venantius: could we fix the indentation on this `:headers` key? you've fixed the indentation for `:body` and now I can't ignore the `:headers` :stuck_out_tongue:

leblowl: thank you, can fix

leblowl:
@ptaoussanis no reason, just wasn't aware I could. It works well & will be a bit quicker that way.

leblowl:
^ something like this. not sure what line of code the comment is referring to here:  https://github.com/leblowl/http-kit/blob/a5cd1be385d4936dfea7afee6b17a5bff8dea519/src/java/org/httpkit/HttpUtils.java#L116... I can move it/reindent it if needs.

ptaoussanis:
> not sure what line of code the comment is referring to here:

Looks like that may need to be one line down, for: `else if (body instanceof ByteBuffer) { return (ByteBuffer) body; }`

ptaoussanis:
Thank you!

PR #258, opened 2016-02-25
Remind the selector to wake up.

### LINKED ISSUE #148: Timeouts under high load with 2.1.17 and 2.1.18
http-kit versions 2.1.17 and 2.1.18 have rare, random timeouts with high load. Version 2.1.16 and earlier work without timeouts. 

Tests are run against this code https://github.com/jsyrjala/http-kit-perf . Tested with [ab](http://httpd.apache.org/docs/2.2/programs/ab.html)

Failing ab runs looks like this. Timeouts don't happen on every run, but run it  a few times, and you should see this behavior.

```
ab -n 10000 -c 100 http://localhost:9876/                      
This is ApacheBench, Version 2.3 <$Revision: 1554214 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
apr_pollset_poll: The timeout specified has expired (70007)
Total of 9994 requests completed
```

venantius:
This commit adds `selector.wakeup()` calls after adding new PendingKeys to the
ConcurrentLinkedList; if we don't do this the main thread blocks on the
`selector.select` call and eventually times out.

This commit also cleans up the polling logic to read a little cleaner, I think.

Resolves #148 

cc @ptaoussanis

### DIFF
--- a/src/java/org/httpkit/server/HttpServer.java
+++ b/src/java/org/httpkit/server/HttpServer.java
@@ -253,9 +253,11 @@ public class HttpServer implements Runnable {
                         selector.wakeup();
                     } else if (!atta.isKeepAlive()) {
                         pending.add(new PendingKey(key, CLOSE_NORMAL));
+                        selector.wakeup();
                     }
                 } catch (IOException e) {
                     pending.add(new PendingKey(key, CLOSE_AWAY));
+                    selector.wakeup();
                 }
             } else {
                 // If has pending write, order should be maintained. (WebSocket)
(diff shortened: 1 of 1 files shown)


ptaoussanis:
You're a champion, thank you! Please feel free to close #148 if you've confirmed that it resolves the issue there.

venantius:
It'll close automatically since the commit message included the magic github text to auto-close issues on merge :)

PR #259, opened 2016-02-26
Reset decoder after starting the close handshake for the websocket se…

### LINKED ISSUE #227: Client sending any byte after starting a close handshake would cause the WS server to crash
I used http-kit for websocket server in three different projects, and each of them crashes every 1-3 days (on average load of 10k connections)

The servers' mem usage was stable, but would increase explosively (for about one minute) before crash.

I sampled several heap dumps, which shows: 

qq 20150826005901

The toWrite linked list of ServerAtta was holding too many nodes, which draws the mem out.

And EVERTY node on the toWrite linked list was holding the same byteBuffer `[-120 2 3 -24 0 0 0 0 0 0 0 0]`, which could be decoded as Websocket: Close Frame.

----------------------------Updated 2015.08.27-------------------------------

The reason has been found

In method  decodeWS, line 140 `src/java/org/httpkit/server/HttpServer.java`, 

```
                } else if (frame instanceof CloseFrame) {
                    handler.clientClose(atta.channel, ((CloseFrame) frame).getStatus());
                    // close the TCP connection after sent
                    atta.keepalive = false;
                    tryWrite(key, WsEncode(WSDecoder.OPCODE_CLOSE, frame.data));
                }
```

The block of frame CloseFrame was missing a call of `atta.decoder.reset();`, which causes the server to send CLOSE Infinitely when a client send any byte after starting a closing handshake, causing the server to run out of memory.

To prove that, We've written a small priece of code to attack our httpkit websocket server, to send a byte immeidately after requesting the WS Close, and it works, the server went crash after a few seconds.

After we added `atta.decoder.reset()` into the closing block, the attack program doesn't crash the server any more

venantius:
…rver

In org.http.server.HttpServer.decodeWs(...), when we receive a CloseFrame, we
don't reset the decoder. This eventually causes the server to run out of memory
as the server just sends CLOSE infinitely.

This commit adds a call to reset the decoder so that we don't get stuck in the
infinite close loop.

Resolves #227

### DIFF
--- a/src/java/org/httpkit/server/HttpServer.java
+++ b/src/java/org/httpkit/server/HttpServer.java
@@ -159,6 +159,7 @@ public class HttpServer implements Runnable {
                     handler.clientClose(atta.channel, ((CloseFrame) frame).getStatus());
                     // close the TCP connection after sent
                     atta.keepalive = false;
+                    atta.decoder.reset();
                     tryWrite(key, WsEncode(WSDecoder.OPCODE_CLOSE, frame.data));
                 }
             } while (buffer.hasRemaining()); // consume all


venantius:
@clyce could you take a look at this and verify that it fixes #227 using the attack code you wrote? (Or could you share the attack code)?

ptaoussanis:
Feel free to merge this (+ anything else) if you're satisfied, thanks!

venantius:
Will do; I'm going to see if we can get some verification from the initial reporters that this fixes their problem before merging.

ptaoussanis:
Since we'll only be cutting an alpha on the next release, I'd suggest just merging this so long which might make it easier to get feedback / catch any problems.

venantius:
Sounds good.

venantius:
Hmm, still encountering timeout issues on tests @mhjort

Issue #203, opened 2015-04-14
Update docs to clarify when error is returned


danielcompton:
Reading the client docs, it wasn't clear that an `error` key is only returned when there is an error in making the request, not when the request returns an error status code e.g. 404 or 500. I think it would be helpful if the docs explained this. I'm happy to put together a PR if the project maintainers agree.

Relates to #175

ptaoussanis:
Sure Daniel, PR welcome!

venantius:
Resolved by #261

PR #261, opened 2016-02-27
Updated documentation on request function regarding to #203

javahippie:
Updated the documentation to clarify the contents of the `:error` key. I am no native speaker so a grammar/spelling check would be nice.


venantius:
This LGTM.

PR #264, opened 2016-02-29
[client] #209 plus some related client fixes

### LINKED ISSUE #209: Redirects wrongly keep query params in place
On redirects the `:query-params` in `opts` are not cleared. You can see it starting [here][1]. So the URI of the GET request after the redirect still contains the query params which is not right. An example network trace looks like this:

```
T 127.0.0.1:44892 -> 127.0.0.1:5001 [AP]
  GET /studies/S.0000/find-form-def?id=F.0000 HTTP/1.1..Accept: application/transit+json..Host: localhost:5001..User-Agent: http-kit/2.0..Accept-Encoding: gzip, deflate..Content-Length: 0....                   
#
T 127.0.0.1:5001 -> 127.0.0.1:44892 [AP]
  HTTP/1.1 301 Moved Permanently..Content-Type: application/transit+json..Vary: Accept..Location: /studies/S.0000/form-defs/F.0000..Access-Control-Allow-Origin: *..Access-Control-Expose-Headers: ETag..Content-Length: 0..Server: http-kit..Date: Tue, 02 Jun 2015 13:35:40 GMT....                                                                                                                                             
#
T 127.0.0.1:44892 -> 127.0.0.1:5001 [AP]
  GET /studies/S.0000/form-defs/F.0000?id=F.0000 HTTP/1.1..Accept: application/transit+json..Host: localhost:5001..User-Agent: http-kit/2.0..Accept-Encoding: gzip, deflate..Content-Length: 0....                
#
```

This will be an easy fix. Should I open a pull request?

[1]: <https://github.com/http-kit/http-kit/blob/2.1.17/src/org/httpkit/client.clj#L162>

jruthe:
- clearing form and query params on 301-303 redirects (fixes #209 and problem when web app redirects after POST)
- SO_KEEPALIVE and TCP_NODELAY on client sockets (this prevents some connection resets when dealing with long running server-side stuff; plus these are a standard browser socket options)

### DIFF
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -345,6 +345,8 @@ public final class HttpClient implements Runnable {
             }
             try {
                 SocketChannel ch = SocketChannel.open();
+                ch.setOption(StandardSocketOptions.SO_KEEPALIVE, Boolean.TRUE);
+                ch.setOption(StandardSocketOptions.TCP_NODELAY, Boolean.TRUE);
                 ch.configureBlocking(false);
                 boolean connected = ch.connect(job.addr);
                 job.isConnected = connected;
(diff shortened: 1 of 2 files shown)


ptaoussanis:
Thanks Jan!

PR #265, opened 2016-03-01
Fix tests in travis

mhjort:
This fixes randomly failing tests from org.httpkit.ws-test in Travis environment. The root cause of failures seems to be the slowness of Travis machines. WebSocketClient in tests uses method `queue.poll(5, TimeUnit.SECONDS). Sometimes 5 seconds is not enough for first message to arrive. ArrayBlockingQueue.poll method does not throw any exception on this but returns null instead. This is a bit misleading because it can also mean that queue is empty.

However, increasing wait time to 10 seconds seems to fix the problem. I ran tests multiple times in Travis and they do not fail anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.iml
 *jar
 .classpath
 .lein-deps-sum
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: clojure
+lein: lein
+script: lein test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,3 @@
 language: clojure
 lein: lein
-script: lein test
+script: lein test :travis
--- a/README.md
+++ b/README.md
@@ -1,34 +1,38 @@
 # HTTP Kit
 
-## High-performance event-driven HTTP client/server for Clojure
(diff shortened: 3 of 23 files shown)


mhjort:
Damn. Now it failed again. Have to dig deeper. This PR should not be merged yet.

mhjort:
Ok. Increased timeout to 30 seconds which helped.

Also I changed the test WebSocketClient implementation to throw an exception if we do not get message within that given timeout. This way it's a lot easier to understand what is failing in tests if we ever hit this problem again.

venantius:
Your branch has gotten pretty messy here; probably a good idea to do a clean rebase of this PR before merging.

venantius:
These changes seem fine otherwise.

venantius:
I think the branch issue isn't your fault; it looks like someone's been re-writing history on master. Not sure if that was you or @ptaoussanis

ptaoussanis:
@venantius That was me, apologies for the confusion. The history since the last tagged release had become a bit unwieldy so squashed some commits to keep things clear + make it easier to revert/bisect while debugging the Travis issue.

@mhjort Thanks a lot for chasing this down! I've merged the PR for you manually. If you happen to run into a merge issue like this again in future, you can just rebase your changes against master.

PR #266, opened 2016-03-02
Include Java source files in the JAR

### LINKED ISSUE #189: Do not exclude Java source files from the JAR
As such there is no source JAR for http-kit and the Java source files are not included in the standard JAR, which makes debugging harder. Please either publish a separate source JAR for every release, or do not exclude the Java source files from the standard JAR.

kumarshantanu:
This PR fixes #189

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.iml
 *jar
 .classpath
 .lein-deps-sum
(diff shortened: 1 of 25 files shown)


ptaoussanis:
Awesome, thank you!

PR #270, opened 2016-03-06
[circleci] Move to CircleCI

venantius:
I've been testing http-kit's tests on CircleCI and they don't seem to be running into the same timeout issues that afflicted us on Travis. They're also free for open source so I just suggest we set that up as our CI server instead of Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,0 @@
-language: clojure
-lein: lein
-script: lein test :travis
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,3 @@
+test:
+  override:
+    - lein test :ci
--- a/project.clj
+++ b/project.clj
@@ -27,7 +27,7 @@
   :jar-exclusions [#"^java.*"] ; exclude the java directory in source path
   :test-selectors
   {:default (complement :benchmark)
-   :travis  (complement #(or (:benchmark %) (:skip-travis %)))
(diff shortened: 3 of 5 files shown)


mhjort:
Good idea. Have you tested if we can also enable ipv6 tests in CircleCI?

venantius:
I did, unfortunately no luck.

mhjort:
Ok. Let's merge this in then.

ptaoussanis:
Thanks, will try handle the setup sometime this week- just need to take a look what the process is.

danielcompton:
It's super simple, just go to circleci.com, sign in with your GitHub account, and select the project to start building.

venantius:
I'd hold off on this for the moment. I've been running a decent number of tests and while the performance is definitely better than Travis I don't yet know if the migration makes sense. We might just be better off disabling the timeout-related tests in CI.

ptaoussanis:
@danielcompton Awesome, thanks

@venantius No problem, will wait on word from you

venantius:
Yeah I mean as you can see here these timeout related tests are just insanely flakey on any cloud provider service: https://circleci.com/gh/venantius/http-kit

I think the better move for us will just be killing the timeout-related tests in CI and keeping track of them otherwise. Another option that would be a good idea (don't remember if I've mentioned this before) would be building in some performance measurement into the tests so that at least we can keep an eye on speed over time and have an audit log for that in our tests.

PR #271, opened 2016-03-08
Revert "[circleci] Move to CircleCI"

venantius:
Reverts http-kit/http-kit#270

I like CircleCI a lot as a product - I think it's the best available - but there's no real reason for us to switch here. Reverting my prior commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: clojure
+lein: lein
+script: lein test :travis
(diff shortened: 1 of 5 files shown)


mhjort:
Yep. Let's stick with Travis. I can take a look on tests when I have time. Maybe there's something to optimize so that they pass even in Travis.

Will merge this

PR #275, opened 2016-03-14
proxy support

davidkazlauskas:
Made basic http proxy support for my project, feel free to merge

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -32,7 +32,11 @@
    :all (fn [_] true)}
 
   :profiles
-  {:test {:java-source-paths ["test/java" "src/java"]}
+  {:test {:java-source-paths ["test/java" "src/java"]
+          :dependencies [[ring/ring-defaults "0.2.0"]
+                         [ring-request-proxy "0.1.3"]
+                         [ring-basic-authentication "1.0.5"]
+                         [org.clojure/data.codec "0.1.0"]]}
    :dev  {:dependencies
           [[junit/junit "4.12"]
            [org.clojure/tools.logging "0.3.1"]
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -212,9 +212,12 @@ public final class HttpClient implements Runnable {
     }
 
     public void exec(String url, RequestConfig cfg, SSLEngine engine, IRespListener cb) {
-        URI uri;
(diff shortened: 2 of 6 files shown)


mhjort:
Would it be possible to add a test case for this feature? I know it's not trivial in this case. However, it would be really nice to get some test coverage on new features.

danielcompton:
`mitmproxy` is one tool I know of that is relatively straightforward to setup as a proxy.

jbristow:
I'm not sure this works.  I created a quick test to see if I could add a proxy server.  It works for clj-http, but not http-kit.

Can someone look at this to verify what I'm seeing? I'm still spinning up, so if someone can point me where I went wrong, please do so! 

https://github.com/jbristow/http-kit/commit/94002782da6caeaa442272a1ccaf6f991c0a9e77

mhjort:
Hi, I checked your test code. And send you comments. Problem was that that proxy was not running in correct port. It worked with clj-http because it seems to ignore the proxy configuration if proxy cannot be connected.

And this new http-kit proxy implementation fails if proxy cannot be connected. Personally I think this is correct way in this case

mhjort:
Also I tested with mitmproxy and verified that this actually works. If you make changes to test code that I mentioned they should pass.

davidkazlauskas:
Thanks for testing this out, sorry for being lazy :(

davidkazlauskas:
I see there's a travis fail, I tried to clone at that commit, first tried lein test org.httpkit.client-proxy-test, then just lein test and got no errors. I see there's a timeout stuff, could travis be retried just in case and can anyone verify that tests pass for them at 680ff1d1a9c7586a0d96104cb5bea997100108ab commit?

mhjort:
I tested locally. Tests pass. Also trying rerun in Travis now. We have been having problems with Travis and async tests

mhjort:
After retry tests pass in Travis. Thanks @davidkazlauskas  and @jbristow 

Only one last thing. Could you be cleanup git history a bit (squash some commits together with meaningful commit msgs). The code looks good now but it consists of 9 different commits. After that this could be merged in.

davidkazlauskas:
Squashed down tests to single commit, still the timeout stuff in travis though.

mhjort:
Thanks. Two retries in Travis helped. I will merge this.

lensgolda:
Please add documentation for proxy support

Issue #280, opened 2016-04-01
New proxy functionality does not handle https properly.


jbristow:
I noticed that the new proxy functionality causes some issues when it tries to call the http proxy with a https request.

`{:opts {:url "https://www.google.com", :method :get, :proxy "http://127.0.0.1:8181"}, :error #<ProtocolException org.httpkit.ProtocolException: No status>}`

I have a PR ready that I think will fix this, but I need some eyes on it to a) figure out if I actually solved the problem and b) make sure I'm testing this right.

jbristow:
The main problem I saw is that when you connect to an http proxy with http-kit, (say it's on my-proxy:80) and then try to connect to https://www.google.com, the request is still marked as https, so this code fails:


```java
// src/java/org/httpkit/client/HttpsRequest.java lines 135-141
    public void recycle(Request old) throws SSLException {
        super.recycle(old);
        this.engine = ((HttpsRequest) old).engine;
        this.handshaken = true;
        wrapRequest(); // prepare for write
    }
```
because the request is a `Request` object and cannot be cast.

jbristow:
It's difficult for me to diagnose exactly because I'm still spinning up on all the mechanics of what's going on in an http request when I do something like this:

```bash
https_proxy="http://localhost:8181s" curl -v -k https://www.google.com/
```

jshaw86:
Was just wondering if there was an update on this? If I can be of help let me know.

jbristow:
@jshaw86 Can you look at [my PR](https://github.com/http-kit/http-kit/pull/281) and see if I'm on the right track?  I'm used to dealing with an apache.httpclient level of abstraction, so this is a little closer to the bone than I'm comfortable with.

jbristow:
Also, looking at the proxies I'm dealing with, it looks like they don't advertise that they're on https at all, they just handle both transparently.

jshaw86:
@jbristow I wanted to make sure I understood the fundamental issue. The way I see it is if we made a https from client -> proxy server that would not be respected without the pull request?

If that's correct then I'll just more or less run the tests and do a sanity check, the if condition though seems reasonable.

jbristow:
Right, when I connect to my egress proxies  using clj-http, I connect via "host:port", and the connection goes through whether I hit `http://www.google.com` or `https://www.google.com`.

jshaw86:
[...]

PR #281, opened 2016-04-01
Adding fix to allow https requests through http-proxy

jbristow:
I need some eyes on it to a) figure out if I actually solved the problem and b) make sure I'm testing this right.

I can't get this to work in mitmproxy so I suspect I missed something.  Either I haven't configured mitmproxy properly, or I've missed something else that needs to be changed.

### DIFF
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -281,7 +281,8 @@ public class HttpClient implements Runnable {
             cb.onThrowable(e);
             return;
         }
-        if ("https".equals(scheme)) {
+        if ((proxyUri == null && "https".equals(scheme))
+            || (proxyUri != null && "https".equals(proxyUri.getScheme()))) {
             if (engine == null) {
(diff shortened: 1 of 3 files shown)


ptaoussanis:
Hi Jon, thanks for your efforts. Could I perhaps ask for a short summary/reminder of what the particular problem is here that you're trying to address?

Are there other ways of addressing this problem? Is it necessary to mod http-kit?

Thanks!

ptaoussanis:
Update: just saw that this relates to #280

jbristow:
The main problem I saw is that when you connect to an http proxy with http-kit, (say it's on my-proxy:80) and then try to connect to https://www.google.com, the request is still marked as https, so this code fails:


```java
// src/java/org/httpkit/client/HttpsRequest.java lines 135-141
    public void recycle(Request old) throws SSLException {
        super.recycle(old);
        this.engine = ((HttpsRequest) old).engine;
        this.handshaken = true;
        wrapRequest(); // prepare for write
    }
```

ptaoussanis:
Closing with #291

PR #284, opened 2016-04-12
Custom error responses for HTTP 413 and 414

skazhy:
This allows returning custom error messages for HTTP 413 and 414. 

example usage:

```clojure
(defn formatter []
  (reify
    HTTPExceptionFormatter
      (errorMessage [_ status e] "custom error message")))

(full.http.server/run-server routes {:exception-formatter (formatter)})
```

### DIFF
--- a/src/java/org/httpkit/DefaultHTTPExceptionFormatter.java
+++ b/src/java/org/httpkit/DefaultHTTPExceptionFormatter.java
@@ -1,0 +1,11 @@
+package org.httpkit;
+
+/**
+ * Default implementation for HTTPExceptionFormatter. Returns the
+ * exception message as-is.
+ */
+public class DefaultHTTPExceptionFormatter implements HTTPExceptionFormatter {
+    public String errorMessage(int status, HTTPException e) {
+        return e.getMessage();
+    }
+}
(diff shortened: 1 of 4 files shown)


venantius:
What was the specific motivation behind needing custom error messages for 413 and 414?

venantius:
Orthogonally - it would be great if you could add some tests to this change.

ptaoussanis:
Hi Kārlis, just a quick ping to clarify/suggest that we confirm that the motivation here does make sense before you go to the effort of adding tests, etc.?

For example- is there a reason why you wouldn't want to control your error messages through a proxy? The usual recommendation would be to run http-kit behind something like nginx for security, static files, http/2, etc. Should also be able to control your error messages that way?

Just hesitant to pick up additional code to the project unless we're pretty sure it's going to be useful to a wide range of folks.

skazhy:
Main reason for this was the ability to format 413 the same way as other exceptions, that are "visible" for the handler.

ptaoussanis:
@skazhy Sorry, could you please clarify what you mean by "visible for the handler"? Also just want to clarify: is this something that can be handled by a reverse proxy (nginx, etc.), or no?

Thanks! Cheers :-)

skazhy:
After going through details on the architecture where this was supposed to be used, it looks like this functionality can live outside of http-kit itself. Thanks for the feedback; closing!

PR #291, opened 2016-05-13
CONNECT Http Tunneling logic

jshaw86:
Our infrastructure requires our clients to tunnel through our HTTP Proxies. I've added logic to allow for tunneling. This is an extension of the @jbristow work on #281.

### DIFF
--- a/src/java/org/httpkit/HttpUtils.java
+++ b/src/java/org/httpkit/HttpUtils.java
@@ -257,6 +257,14 @@ public class HttpUtils {
         }
         return host;
     }
+    
+    public static String getProxyHost(URI uri){
+        if (uri.getPort() == -1){
+            return uri.getHost();
+        }
+        
+        return uri.getHost() + ":" + uri.getPort();
+    }
 
     public static InetSocketAddress getServerAddr(URI uri) throws UnknownHostException {
         InetAddress host = getByName(uri.getHost());
--- a/test/org/httpkit/ring_request_proxy.clj
+++ b/test/org/httpkit/ring_request_proxy.clj
@@ -1,0 +1,71 @@
+(ns org.httpkit.ring-request-proxy
+  (:require [clj-http.client :as clj-http]))
+
+;;
+;; code from: https://github.com/FundingCircle/ring-request-proxy
+;; Copyright (c) 2015, Funding Circle
+;; All rights reserved.
+;;
+;; Redistribution and use in source and binary forms, with or without modification,
+;; are permitted provided that the following conditions are met:
+;;
+;; 1. Redistributions of source code must retain the above copyright notice, this
+;;    list of conditions and the following disclaimer.
+;;
+;; 2. Redistributions in binary form must reproduce the above copyright notice,
+;;    this list of conditions and the following disclaimer in the documentation
+;;    and/or other materials provided with the distribution.
+;;
+;; 3. Neither the name of the copyright holder nor the names of its contributors
+;;    may be used to endorse or promote products derived from this software without
+;;    specific prior written permission.
+;;
+;; THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+;; ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+;; WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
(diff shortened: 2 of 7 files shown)


jshaw86:
I can't repro this on my local, the error also seems unrelated to the patch.

```
$ lein test :only org.httpkit.ws-test/test-binary-frame

lein test org.httpkit.ws-test

Ran 1 tests containing 5 assertions.
0 failures, 0 errors.
```

jbristow:
I think this is closer to the (admittedly vague) standard.  Adding some comments in-line.

src/java/org/httpkit/HttpUtils.java
```diff
@@ -257,6 +257,14 @@ public static String getHost(URI uri) {
         }
         return host;
     }
+    
+    public static String getProxyHost(URI uri){
```
jbristow: I think this function is telling us that `:proxy "http://some.proxy:8080"` is not a good way to express the proxy.

Should we recommend changing this to `:proxy-host "some.proxy" :proxy-port 8080`

test/org/httpkit/ring_request_proxy.clj
```diff
@@ -0,0 +1,71 @@
+(ns org.httpkit.ring-request-proxy
```
jbristow: I think this is unused, delete?

src/java/org/httpkit/HttpUtils.java
jshaw86: I like that because it's closer to clj-http I think but we would also have to put in :proxy-scheme (default to http?) might get verbose from a config perspective. thoughts?

test/org/httpkit/ring_request_proxy.clj
jshaw86: https://github.com/pubnub/http-kit/blob/6c00b51ee447d1b00ca614c21b54b9208b5b87f9/test/org/httpkit/client_proxy_test.clj#L20 it's being used there

src/java/org/httpkit/HttpUtils.java
jbristow: This is also a good point.  Does clj-http infer the scheme for a request-level proxy override from the request uri scheme?

Also, should I add a feature-request for http-kit to respect the [Java "standard" networking properties](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html)?

jshaw86: It looks like they default to just HTTP https://github.com/dakrone/clj-http/blob/master/src/clj_http/core.clj#L117. Unless they modify the scheme of the host later on, but they aren't using the 3 arity form of HttpHost that allows you to set the scheme. https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/org/apache/http/HttpHost.html

jshaw86:
still can't replicate these websocket errors, must be something on travis end

jshaw86:
@jbristow here are the changes you requested

jshaw86:
who is in charge of merging pull requests? any update on getting this merged?

ptaoussanis:
Hi there, will try take a look this weekend - just travelling atm. What's the current status on this? You're both happy with the current PR? It's stable + tested? This also resolves #280 and #281, is that correct?

> I like that because it's closer to clj-http I think but we would also have to put in :proxy-scheme (default to http?) might get verbose from a config perspective. thoughts?

What was the conclusion on this? How is the proxy scheme specified as of this PR?

jshaw86:
Hey Peter,
Thanks for your response. @jbristow wrote tests for the proxy code that I modified. So we are currently testing and passing both http and https tests through a proxy. The current embedded proxy servers in the test though don't support the CONNECT verb which is enabled when setting tunnel? = true. I did some research and this is fairly complicated to setup for E2E testing. I could potentially mock off the request and just verify the payload being sent for tunnel? true but the current coverage imo should be sufficient to verify the behavior.

The proxy scheme is given in the proxy-host here is an example:
```
{:proxy-host "http(s)://127.0.0.1"
:proxy-port 3128}
```

I'm currently running a custom jar with the PR in a production system with a squid proxy between the http-kit client and interwebs for the last week or so and haven't seen issues.

I still can't explain the WS failure in on travis though as I can't replicate it locally.

ptaoussanis:
@jshaw86 Okay, thanks for the clarifications- sounds good.

Just to confirm: this PR also resolves #280 and #281, is that correct?

> I still can't explain the WS failure in on travis though as I can't replicate it locally.

There's known issues with the Travis tests; would happily also accept a PR that just disables them for now to avoid further confusion.

jshaw86:
> Just to confirm: this PR also resolves #280 and #281, is that correct?

I would say yes.

ptaoussanis:
Merged, thanks!

PR #293, opened 2016-05-19
add support for a max connections in HttpClient

ryfow:
Is there interest in pursuing a max-connections setting in the http-kit client? I'm up for working on it with someone.

Right now, it's pretty easy run yourself out of file handles or to tip over a remote server. I resorted to using a thread pool and synchronous requests in one project, but doing that made me sad.

This is one implementation idea where we just keep a counter of numConnections.

I obviously don't know the codebase, so I could use help from a maintainer. I'm also interested in hearing people's ideas for writing good tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,0 @@
-language: clojure
-lein: lein
-script: lein test :travis
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: clojure
+lein: lein
+script: lein test :travis
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,35 @@
+# How to Contribute to http-kit
+
+Interested in making http-kit better? Radical. Let's talk about how you can get
+the most mileage out of it.
+
+## Be Respectful
+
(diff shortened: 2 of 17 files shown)


ptaoussanis:
@ryfow Hi Ryan, thanks for this. Is there a reason you wouldn't want to just rate limit your client requests Clojure-side?

ryfow:
Good point. Right now I'm using a Java Semaphore outside the client with something like this:

```clojure
(let [sem (Semaphore. 100)]
  (doseq [term search-terms]
    (.acquire sem)
    (client/get "http://example.org"
                {:query-params {:q term}}
                (fn [resp]
                  (.release sem)
                  ;; process results
                  ))))
```
This works, but it seems like something the client should be hiding from me.

I could follow this pattern in `HttpClient.java` or in `client.clj`. Do you have a preference between the two? Seems that we could hide the semaphore in HttpClient, but probably have the library user pass the actual semaphore to `client/request` in the same way they can pass `worker-pool`.

Do you have a better way of rate limiting requests from the clojure side?

ptaoussanis:
> This works, but it seems like something the client should be hiding from me

Okay, let's try it out. PR looks reasonable to me, have merged. Cheers :-)

PR #297, opened 2016-06-14
Multipart tweaks

vincentjames501:
Allows setting the content-type for each multipart entity.

I also set it to allow byte arrays in a separate commit. This caught me off guard as no errors were ever thrown - just nothing would happen when posting them and it seems to be a pretty common thing people would want to do.


ptaoussanis:
Looks good Vincent, much thanks!

PR #300, opened 2016-07-20
Added a getCount method to BytesInputStream

MysteryMachine:
At my company, we use the length of inputs to do load estimation on our backend. As of right now, there is no easy way to access the length of a BytesInputStream other than calling toString on it, and parsing out the returned length. 

This PR provides an accessor to count in BytesInputStream. Since it is already exposed in toString, I figured it would not be a problem.

I'm eager to get this in, let me know if there are any changes you'd like! Or, if you'd prefer some alternative approach, let me know and I'll do things your way.

### DIFF
--- a/src/java/org/httpkit/BytesInputStream.java
+++ b/src/java/org/httpkit/BytesInputStream.java
@@ -80,4 +80,8 @@ public class BytesInputStream extends InputStream {
     public void reset() {
         pos = mark;
     }
+
+    public int getCount() {
+        return count;
+    }
 }


venantius:
This seems fine to me. Thanks for submitting.

ptaoussanis:
Thanks both!

ptaoussanis:
@venantius May I request that we merge as with the [previous style](https://github.com/http-kit/http-kit/commits/master) when possible? This project gets many small contributions, and having everything indexed on master saves me a lot of time if we need to do bisecting, or when cutting releases, etc. 

Cheers! :-)

venantius:
@ptaoussanis of course, sorry about that. It occurred to me that I should have used your format almost immediately after I merged it :(.

ptaoussanis:
Zero problem. Appreciate all your help :-)

Issue #302, opened 2016-08-11
:proxy-host param is misleading


taso42:
it's named `:proxy-host` but it's really the proxy scheme and proxy host.

consider a single param, `:proxy-url` in place of `:proxy-host` and `:proxy-port`; or introducing a 3rd param `:proxy-scheme`

taso42:
Alternatively (or additionally) honor the Java system properties
`http.proxyHost`, `http.proxyPort`, `https.proxyHost`, `https.proxyPort`

PR #303, opened 2016-08-12
replace :proxy-host and :proxy-port parameters with a singular proxy-url parameter

taso42:
See #302

### DIFF
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -224,8 +224,8 @@ public class HttpClient implements Runnable {
         URI uri,proxyUri = null;
         try {
             uri = new URI(url);
-            if (cfg.proxy_host != null) {
-                proxyUri = new URI(String.format("%s:%03d", cfg.proxy_host, cfg.proxy_port));
+            if (cfg.proxy_url != null) {
+                proxyUri = new URI(cfg.proxy_url);
             }
         } catch (URISyntaxException e) {
             cb.onThrowable(e);
--- a/src/java/org/httpkit/client/RequestConfig.java
+++ b/src/java/org/httpkit/client/RequestConfig.java
@@ -12,24 +12,22 @@ public class RequestConfig {
     final Object body;
     final Map<String, Object> headers;
     final HttpMethod method;
-    final String proxy_host;
-    final int proxy_port;
+    final String proxy_url;
     final boolean tunnel;
 
     public RequestConfig(HttpMethod method, Map<String, Object> headers, Object body,
-                         int timeoutMs, int keepAliveMs, String proxy_host, int proxy_port, 
+                         int timeoutMs, int keepAliveMs, String proxy_url,
                          boolean tunnel) {
         this.timeout = timeoutMs;
         this.keepAlive = keepAliveMs;
         this.headers = headers;
         this.body = body;
         this.method = method;
-        this.proxy_host = proxy_host;
-        this.proxy_port = proxy_port;
+        this.proxy_url = proxy_url;
         this.tunnel = tunnel;
     }
 
     public RequestConfig() { // for easy test only
-        this(HttpMethod.GET, null, null, 40000, -1, null,-1,false);
+        this(HttpMethod.GET, null, null, 40000, -1, null, false);
     }
 }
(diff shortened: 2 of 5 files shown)


taso42:
@davidkazlauskas, @jbristow - i thought i should tag you since you worked on the initial proxy support.

ptaoussanis:
@taso42 Thanks for this. As a general suggestion, please wait on confirmation of a change proposed in an issue before submitting PRs.

Only had time to skim this, but an initial objection would be introducing a breaking change. Is there a strong reason the PR cannot provide backwards compatibility with the previous API?

Otherwise, will leave it to David and Jon to provide feedback.

Thanks!

davidkazlauskas:
I'm sorry, I think this is silly. This would introduce a breaking change... Because... Ugh... It may be unclear in 1 out of 10000000 times that you can specify proxy port in proxy url and we need to have two parameters explicitly every time? Is there some url/port combination that could not be specified in current single parameter?

Yeah, I didn't think of the naming that much but I believe it's close enough, it is not :banana-host :monkey-port or :rfc-791-address so I don't think people are that stupid and could not figure out how to use this parameter. If people want to find out how to use http proxy they should lookup docs and docs should give the unambiguous answer instead of breaking because naming is "somewhat unclose to the truth".

Last time I checked there are bazillions of http servers where you can specify the ip on which to listen and port at the same time, usually, I didn't even notice how the parameter name is called, nobody died.

ptaoussanis:
@davidkazlauskas Please watch the tone here, thanks. And I'm not sure I understand your response?:

> Is there some url/port combination that could not be specified in current single parameter?

The current API requires 2 parameters, this PR adjusts to a single parameter. Sorry, am I missing something obvious?

The general objective seems sensible enough to me, my only request would be that backward compatibility be added.

davidkazlauskas:
> I'm sorry, I think this is silly. This would introduce a breaking change... Because... Ugh... It may be unclear in 1 out of 10000000 times that you can specify proxy port in proxy url and we need to have two parameters explicitly every time? Is there some url/port combination that could not be specified in current single parameter?
> 
> Yeah, I didn't think of the naming that much but I believe it's close enough, it is not :banana-host :monkey-port or :rfc-791-address so I don't think people are that stupid and could not figure out how to use this parameter. If people want to find out how to use http proxy they should lookup docs and docs should give the unambiguous answer instead of breaking because naming is "somewhat unclose to the truth".
> 
> Last time I checked there are bazillions of http servers where you can specify the ip on which to listen and port at the same time, usually, I didn't even notice how the parameter name is called, nobody died.

Sorry, didn't understand PR fully, I thought it was already one URL already, yeah, then I guess this would be appropriate but backwards compatibility should be kept.

ptaoussanis:
No problem, thanks for clarifying.

@taso42 So to update: happy to take a look at a PR adjusted to include backwards compatibility for the current API, thanks!

taso42:
Fair point about backward compatibility.  I will adjust accordingly against this PR.  (Or, if you prefer, go please go ahead and close this PR and I will resubmit when it's done in a back compat way)

taso42:
the last 2 commits now restore backward compatability

jbristow:
We only split it into two to be more consistent with clj-http and to sort-of emulate the way HttpComponents reads the System properties 'http.proxyHost' and 'http.proxyPort'. (or https.*)

As long as backwards compatibility is maintained, I'm cool.  Though the tests should illustrate the backwards compatibility as well as the new functionality.

taso42:
@jbristow https://github.com/http-kit/http-kit/pull/303/commits/028469ed98a812f8809dcab4401f4778e6d1a136 added the backward compat test.  Please let me know if you find it sufficient.

I am currently debating with myself whether the http.proxyHost, http.proxyPort system properties should also be honored while we're at it.  But that can be a totally separate PR.   If anyone has thoughts on that, please comment on #302.  And if there is interest, perhaps it should ultimately be spun off into a separate issue.

davidkazlauskas:
Yeah, I don't see any worth in using java system properties either. I'm good if you ask me.

ptaoussanis:
> But that can be a totally separate PR. If anyone has thoughts on that, please comment on #302. And if there is interest, perhaps it should ultimately be spun off into a separate issue.

Can't comment on the level of interest, but I would please request a separate PR if you wanted to submit for that.

> Can this change be merged as it now stands?

Sorry, haven't had an opportunity to look over the PR properly yet. Just have my hands full atm, will try get this merged w/in the next 2 weeks.

ptaoussanis:
Merged, thanks again!

Issue #187, opened 2015-01-29
Client feature request: add SNI support


timreinke:
Hi,

I ran into this and I'm not sure how to debug.  I don't think it's the trust store - I dug into that.  Are you able to reproduce the below?

My environment is http-kit 2.1.18, Clojure 1.6, Java 1.7, OS X 10.10.2.

``` clojure
;; slurp works
;; sunlight> (slurp "https://congress.api.sunlightfoundation.com")
"{\"thank_you\":\"for using https\",\"status\":200,\"message\":\"I live!\",\"documentation\":\"https://sunlightlabs.github.io/congress/\",\"code\":\"https://github.com/sunlightlabs/congress\",\"report_bugs\":\"https://github.com/sunlightlabs/congress/issues\",\"more_apis\":\"http://sunlightfoundation.com/api/\"}"

;; http-kit doesn't
;; sunlight> @(http/request {:url "https://congress.api.sunlightfoundation.com", :method :get})
{:opts {:method :get, :url "https://congress.api.sunlightfoundation.com"}, :error #<SSLException javax.net.ssl.SSLException: Received fatal alert: handshake_failure>}

;; google works
;; sunlight> @(http/request {:url "https://google.com", :method :get})
"this request works"
```

nickzam:
That's because http-kit client does not support SNI.
with SNI:
```
$ openssl s_client -connect congress.api.sunlightfoundation.com:443  -servername 'congress.api.sunlightfoundation.com'
CONNECTED(00000003)
depth=3 /C=SE/O=AddTrust AB/OU=AddTrust External TTP Network/CN=AddTrust External CA Root
verify error:num=19:self signed certificate in certificate chain
verify return:0
< output suppressed >
```
and without SNI:
```
$ openssl s_client -connect congress.api.sunlightfoundation.com:443                                                   
CONNECTED(00000003)
26577:error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure:/SourceCache/OpenSSL098/OpenSSL098-52.8.3/src/ssl/s23_clnt.c:593:
```

ptaoussanis:
Closing this issue for inactivity, please feel free to reopen if anyone has new input to contribute.

stijnopheide:
Hi @ptaoussanis, I think this is still a valid feature request. The only work-around available right now is to set the request options to ```:insecure? true```.

ptaoussanis:
[...]

PR #311, opened 2016-10-14
Minimal fix to add SNI support to client

sattvik:
Here is an initial effort at resolving #187.  I welcome any feedback, since I
am not very familiar with the http-kit code or the async NIO library.

This works for my particular situation, but there are a couple of caveats:
1. This code is Java 8 only.  If there is a need to support older versions of
   Java, this logic will need to be wrapped.
2. It’s possible that a bad hostname could result in an
   `IllegalArgumentException`.  Is it better for that to be allowed to bubble
   up or should it be swallowed and the SNI support disabled.

Lastly, is this the righ insertion point for adding the SNI support?  I do
not know if there could be any race conditions or shared state that could
result from this.

### DIFF
--- a/src/java/org/httpkit/HttpMethod.java
+++ b/src/java/org/httpkit/HttpMethod.java
@@ -11,7 +11,7 @@ public enum HttpMethod {
     CONNECT(intern("connect")), PATCH(intern("patch")), PROPFIND(intern("propfind")),
 	PROPPATCH(intern("proppatch")), LOCK(intern("lock")), UNLOCK(intern("unlock")),
 	REPORT(intern("report")), ACL(intern("acl")), MOVE(intern("move")), 
-	COPY(intern("copy"));
+	COPY(intern("copy")), MKCOL(intern("mkcol"));
 
     public final Keyword KEY;
 
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -3,9 +3,12 @@ package org.httpkit.client;
 import org.httpkit.*;
 import org.httpkit.ProtocolException;
 
+import javax.net.ssl.SNIHostName;
+import javax.net.ssl.SNIServerName;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLException;
+import javax.net.ssl.SSLParameters;
 import java.io.IOException;
 import java.net.*;
 import java.nio.ByteBuffer;
@@ -13,7 +16,9 @@ import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.nio.channels.SocketChannel;
 import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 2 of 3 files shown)


ptaoussanis:
@sattvik Hi Daniel, thanks a lot for this. Looks promising, and definitely keen to get SNI support in.

Will try schedule a look w/in the next week or two. 

Otherwise, feedback from anyone would be welcome in the meantime.

juhakaremo:
Thank you, Daniel.

I tested this patch and as far as I can tell this patch fixes the SNI issue. Without this fix we are unable to connect to our AWS deployed services using http-kit. 

I will not comment on the actual code changes since I'm not that familiar with http-kit code base or SNI.

stevenschlansker:
hi @ptaoussanis we just ran into a production incident where missing this feature was a contributing factor.  So our feedback is, "this feature is important!" 😄 The code looks reasonable to me (given not really knowing all that much about `http-kit` or Clojure)

soumyapaul30:
What is the SLA to get this fix available for us to use? This is very much in need at the moment.

taffowl:
Also just hit this issue also whilst trying to do a get request to aws apigateway using `http-kit`

ptaoussanis:
@sattvik: Hi Daniel, I'm really sorry for taking so long to look at this - really do appreciate your efforts on this!

> It’s possible that a bad hostname could result in an IllegalArgumentException. Is it better for that to be allowed to bubble up or should it be swallowed and the SNI support disabled.

I believe surfacing the exception is probably better.

> This code is Java 8 only. If there is a need to support older versions of Java, this logic will need to be wrapped.

So this may unfortunately be a blocker since I'd rather we not break JVM compatibility if we can avoid it. 

Is there any way I may be able to request a wrapper so that only those wishing to use the new SNI support require Java >= 8? I.e. we retain support for older JVM versions for those not needing SNI?

May also request a unit test if possible to exercise the SNI support?

If you (or someone else) can get a modified PR up, I'll make sure to get it merged right away!

Again, sincere thanks.

kumarshantanu:
> Is there any way I may be able to request a wrapper so that only those wishing to use the new SNI support require Java >= 8? I.e. we retain support for older JVM versions for those not needing SNI?

I have submitted https://github.com/http-kit/http-kit/pull/335 to decouple Java 8 changes from SSLEngine customization in the client.

ptaoussanis:
Hi @kumarshantanu, thank you! Will try look at your PR at the weekend!

tdg5:
Is there anything folks in the community can do to help get this patch merged sooner?

- [x] It seems like @kumarshantanu has [taken a stab at addressing the Java 8 dependency](https://github.com/http-kit/http-kit/pull/335)
- [x] From what I can tell the concern regarding the `IllegalArgumentException` when a bad hostname is encountered doesn't require any additional work as the current behavior is the desirable behavior.
- [ ] A unit test

Anything else outstanding that would help move this along?

ptaoussanis:
@tdg5 Hi Danny, sincere apologies for the slow response- will try my best to get this merged tonight after work!

ptaoussanis:
Merged! Released on Clojars as `[http-kit "2.3.0-alpha3"]` :-)

icambron:
@ptaoussanis I'm a bit confused. From the discussion above, it sounds like you were rejecting this change for being Java 8-only and were intending to merge #335 to let the caller set up SNI and inject it. But you merged this one and not that one. Is this a) a mistake and you meant to merge the other one or b) a change in your decision re: Java 8?

ptaoussanis:
@icambron Hi Isaac, thanks for checking - option a): this was indeed a mistake on my part. Had only skimmed it and misunderstood the comment by https://github.com/http-kit/http-kit/pull/311#issuecomment-324423610.

Will try take a proper look at this this weekend. Apologies for the confusion.

ptaoussanis:
Okay, just cut https://github.com/http-kit/http-kit/releases/tag/2.3.0-alpha4!

PR #330, opened 2017-03-07
don't override Date header if it's been set by application

ryfow:
If an application developer explicitly sets a Date header, I think http-kit should trust that they meant to.


ptaoussanis:
Great, thanks for the PR Ryan! Apologies for the delay merging.

PR #332, opened 2017-04-03
Do not respond to unsolicited pong frames

### LINKED ISSUE #322: Websockets with Internet Explorer cause intense load
I am using http-kit and Sente to communicate with clients (Chrome and Internet Explorer) via websockets. My application exhibits a high CPU load on the server with little user activity.

The reason for this turned out to be idle IE browsers sending unsolicited pong frames to the server. The server needs to respond to this pong frame as otherwise IE will send an invalid message and cause the connection to terminate.

http-kit contains a fix for this in https://github.com/http-kit/http-kit/pull/134. This fix instructs http-kit to respond with a ping frame.

However, that response causes the client to send another pong frame, as required by the websocket spec. This is understood as another unsolicited pong frame by the server. This will be a start of an intense match between the server and the client.

I prepared a proto which responds with a pong frame. This seems to satisfy the client and produces only occasional messaging.

https://github.com/mikkosuonio/http-kit/tree/mikkosuonio-http-kit-ie-ping-pong

Would this response be appropriate for an unsolicited pong frame sent by IE?

mikkosuonio:
Internet Explorer and MS Edge send unsolicited pong frames. Responding to
those with ping frames starts a match.

This fixes #322.

### DIFF
--- a/src/java/org/httpkit/server/HttpServer.java
+++ b/src/java/org/httpkit/server/HttpServer.java
@@ -153,8 +153,8 @@ public class HttpServer implements Runnable {
                     atta.decoder.reset();
                     tryWrite(key, WsEncode(WSDecoder.OPCODE_PONG, frame.data));
                 } else if (frame instanceof PongFrame) {
+                    // ignored as unsolicited pong frame from client
                     atta.decoder.reset();
-                    tryWrite(key, WsEncode(WSDecoder.OPCODE_PING, frame.data));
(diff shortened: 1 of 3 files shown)


ptaoussanis:
Brilliant, thank you Mikko! Sorry for the delay getting this merged.

davewo:
@ptaoussanis would you be open to accepting a backport of this fix into the 2.2.0 line? Or, are we getting close a 2.3.0 release :-) ?

Issue #323, opened 2017-01-29
Metrics event instrumentation


kumarshantanu:
Http-Kit knows about several events that would be useful to emit and aggregate outside of the server. Examples include: latency of the hostname resolution in http-client, the times when the server responded with HTTP 503 because the request queue was full etc.

I have been maintaining a fork for quite some time that allows one to _optionally_ pass event logger as argument to capture various events. The overhead when one doesn't want any logging is practically zero due to JVM JIT's empty-method-body elimination. This fork has seen over 6 months of production use so far. It has helped troubleshoot infrastructure bug and server behavior under load via metrics logging/aggregation. (I would be happy to offer the changes I have maintained as a pull request, and to make any further changes needed to make the PR suitable for merging.) The URL of the fork diff is below:
https://github.com/http-kit/http-kit/compare/master...kumarshantanu:logging?expand=1

ptaoussanis:
Hi @kumarshantanu, sorry for the crazy delay responding!

Would definitely be open to seeing a PR for this, thank you!

ptaoussanis:
Somehow missed closing this earlier, Shantanu's PR was merged in 2017 👍

PR #333, opened 2017-04-18
Logging instrumentation

kumarshantanu:
This PR addresses https://github.com/http-kit/http-kit/issues/323 with the following changes:

- Enables passing loggers (Java interface, default implementation is no-op) as argument
- Logger accepts event name
- Event names are customizable (defaults can be overridden)
- The `AddressFinder` interface is to capture the name resolution latency in the client

This PR also fixes the issue where using a custom client object (in `client.clj`) still dereferences the "default client" object, thereby starting up a redundant thread.

I have verified that all tests pass with the changes. I have also released `[org.clojars.kumarshantanu/http-kit "2.3.0-alpha2-logging4"]` in case anybody wants to test a build.


ptaoussanis:
Awesome, much thanks! Merged :-)

PR #341, opened 2017-06-23
Increase the max-line default to conform to nginx defaults

### LINKED ISSUE #196: Request header max line length small by default
When the server parses request lines, it refuses to parse header lines with a length < max_length (see [here](https://github.com/http-kit/http-kit/blob/master/src/java/org/httpkit/LineReader.java#L32)). I ran into this issue because I kept seeing errors of this type: `414 Request-URI Too Long`.

The documentation covers this. The problem goes away if you set `:max-line` to a larger value as documented [here](http://www.http-kit.org/server.html).

Here's a test script to trigger the behavior:

````
curl -i "http://localhost:3000/route" -H "Cookie: $(perl -e 'print "A" x 4100 . "\n";')" 
````

It's actually common for browsers to send large cookies in a single header line. I'm looking at a request sent by Chrome with a "Cookie:" line with >5000 bytes.

I see two problems with http-kit here:

1. Most importantly, the default is low. Nginx uses a default max of 8192 bytes: http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers. Apache uses a default max of 8190 bytes: http://httpd.apache.org/docs/2.2/mod/core.html#limitrequestline

   I'd suggest using 8192 bytes as a default to match the behavior of other common httpds.

2. The message "Request-URI Too Long" is misleading (the actual request uri is fine, it's the header line length that is exceeded). Apache returns the more informative `HTTP/1.1 400 Bad Request`; nginx does the same.

Kjir:
Fixes #196

### DIFF
--- a/src/org/httpkit/server.clj
+++ b/src/org/httpkit/server.clj
@@ -42,7 +42,7 @@
               queue-size 20480
               max-body   8388608
               max-ws     4194304
-              max-line   4096
+              max-line   8192
               proxy-protocol :disable
               worker-name-prefix "worker-"}}]]
 
--- a/test/org/httpkit/server_test.clj
+++ b/test/org/httpkit/server_test.clj
@@ -289,12 +289,12 @@
 
 (deftest test-uri-len
   (doseq [_ (range 0 50)]
-    (let [path (str "/abc" (subs const-string 0 (rand-int 4000)))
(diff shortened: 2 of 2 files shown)


ptaoussanis:
Merging now, thanks!

PR #360, opened 2017-12-07
Error while processing chunked responses split between CRLF characters

rkaippully:
Fixes issue #3. Handle cases where the bytes read may not contain both CR and
LF characters in the same packet.


rkaippully:
Apologies, this is not ready yet. I'll send a new PR with correct details.

PR #362, opened 2017-12-08
Error while processing chunked responses split between CRLF characters

### LINKED ISSUE #361: [bug] Response of one request gets delivered to another request
version: `2.3.0-alpha4`

I encountered this interesting bug in http-kit client. The symptom was that a few times a day some of our POST requests to a server running Tomcat will return a response that is intended for another request executed concurrently sharing the same http client. For e.g. if requests R1 and R2 are running concurrently in two threads and share the same `HTTPClient`, R2 gets the response of R1 and R1 eventually times out. When we inspected the logs from the server, it showed that the server sent both the responses correctly.

After quite some debugging it turned out that there is a subtle bug in the parsing logic for chunked responses that can cause this. I have a fix for this undergoing testing. If you are interested I can send a pull request from my fork (https://github.com/rkaippully/http-kit) as soon as I complete testing.

This sounds similar to the issue reported in #338, but I am not sure if it is the same bug. That issue seem to be affecting the server, this issue is for the client.

rkaippully:
This fixes #361. Here is an explanation of the bug: https://github.com/http-kit/http-kit/issues/361#issuecomment-350266158

The fix is two fold:

1. Fix the logic of the decoder so that it works across ByteBuffer boundaries.
2. Even if somehow the decoder gets to the `ALL_READ` state more than once, ensure that `HTTPClient.doRead()` do not create multiple `PersistentConn` objects for the same `SelectionKey`.

I have added unit test demonstrating the bug. This fix has been running in our production environment for some time now without any glitches.


ptaoussanis:
Great work, thank you- merging manually now!

PR #363, opened 2017-12-14
added global client request interceptor

mattjones753:
A bit of background for the change: We are looking at using [correlation ids](https://blog.rapid7.com/2016/12/23/the-value-of-correlation-ids/) across a number of clojure services. This means we needed a means of propogating a correlation ID as a header from incoming requests to all outgoing calls. We wanted to be able to have a global means of handling this without having to change all outgoing client calls to add the header. I thought a global request interceptor would allow this and also be a bit more flexible to allow others to do similar things.

It will still allow users to override the result of the interceptor on a per-request basis.

The change was a bit speculative so if it doesn't fit with the approach of this library or if any changes need to be made, please let me know!

### DIFF
--- a/src/org/httpkit/client.clj
+++ b/src/org/httpkit/client.clj
@@ -23,6 +23,13 @@
                      (str (first basic-auth) ":" (second basic-auth)))]
     (str "Basic " (base64-encode (utf8-bytes basic-auth)))))
 
+(def request-interceptor (atom nil))
+
+(defn set-request-interceptor
+  "sets a global interceptor for outgoing requests"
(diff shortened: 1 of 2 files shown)


ptaoussanis:
Hi Matt,

Thanks a lot for the PR (and for your patience)! So I looked over this and gave it some thought. To be honest, I don't think this is right for http-kit core.

Doing request preparation is already quite easy at the application level, and doing it there has the benefit of minimizing complexity and maintenance burden inside http-kit core.

Hope that makes sense?

In any case, was happy to consider this- thanks again!

Issue #355, opened 2017-10-19
Concurrent modification exception on server shutdown


cnuernber:
(Using version 2.2.0, but I checked and the bug appears to still exist):

```clojure
Error during http server shutdown:
#error {
 :cause nil
 :via
 [{:type java.util.ConcurrentModificationException
   :message nil
   :at [java.util.HashMap$HashIterator nextNode "HashMap.java" 1437]}]
 :trace
 [[java.util.HashMap$HashIterator nextNode "HashMap.java" 1437]
  [java.util.HashMap$KeyIterator next "HashMap.java" 1461]
  [java.util.Collections$UnmodifiableCollection$1 next "Collections.java" 1042]
  [org.httpkit.server.HttpServer stop "HttpServer.java" 334]
  [org.httpkit.server$run_server$stop_server__6415 doInvoke "server.clj" 56]
```

From the documentation on java.io.Selector (https://docs.oracle.com/javase/7/docs/api/java/nio/channels/Selector.html):

```
Concurrency

Selectors are themselves safe for use by multiple concurrent threads; their key sets, however, are not.

The selection operations synchronize on the selector itself, on the key set, and on the selected-key set, in that order. They also synchronize on the cancelled-key set during steps (1) and (3) above.

Changes made to the interest sets of a selector's keys while a selection operation is in progress have no effect upon that operation; they will be seen by the next selection operation.

Keys may be cancelled and channels may be closed at any time. Hence the presence of a key in one or more of a selector's key sets does not imply that the key is valid or that its channel is open. Application code should be careful to synchronize and check these conditions as necessary if there is any possibility that another thread will cancel a key or close a channel.

A thread blocked in one of the select() or select(long) methods may be interrupted by some other thread in one of three ways:

By invoking the selector's wakeup method,

By invoking the selector's close method, or

By invoking the blocked thread's interrupt method, in which case its interrupt status will be set and the selector's wakeup method will be invoked.

The close method synchronizes on the selector and all three key sets in the same order as in a selection operation.

A selector's key and selected-key sets are not, in general, safe for use by multiple concurrent threads. If such a thread might modify one of these sets directly then access should be controlled by synchronizing on the set itself. The iterators returned by these sets' iterator methods are fail-fast: If the set is modified after the iterator is created, in any way except by invoking the iterator's own remove method, then a ConcurrentModificationException will be thrown.
```

[...]

PR #369, opened 2018-03-08
[#355] workaround for missing safe key iteration method when stopping

smee:
As the bug ticket #355 correctly stated, there is no safe way to iterate over `selector.keys()` without any `ConcurrentModificationException`  as soon as we close any client connection. Since we can rely on this behaviour according to the official documentation in https://docs.oracle.com/javase/7/docs/api/java/nio/channels/Selector.html we just retry the closing operation until we successfully closed all connections.


ptaoussanis:
Hi Steffen, this looks good to me- thanks!

ptaoussanis:
Merged manually :+1:

Issue #356, opened 2017-10-24
Java 9 deps problem `javax.xml.bind.DatatypeConverter` missing


claj:
# Problem

When upgrading my JVM to 9.0.1 there is a problem with a javax.xml.bind.DatatypeConverter class missing. This makes http-kit client unusable on JVM 9 which I consider serious.''

# http-kit version
`[http-kit "2.2.0"]`, `[http-kit "2.3.0-alpha4"]`

# Java version
java version "9.0.1"
Java(TM) SE Runtime Environment (build 9.0.1+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.1+11, mixed mode)

# Stack trace
`Caused by: java.lang.ClassNotFoundException: javax.xml.bind.DatatypeConverter, compiling:(org/httpkit/client.clj:1:1)
	at clojure.lang.Compiler.load(Compiler.java:7526)
	at clojure.lang.RT.loadResourceScript(RT.java:379)
	at clojure.lang.RT.loadResourceScript(RT.java:370)
	at clojure.lang.RT.load(RT.java:460)
	at clojure.lang.RT.load(RT.java:426)
	at clojure.core$load$fn__6550.invoke(core.clj:6051)
	at clojure.core$load.invokeStatic(core.clj:6050)
	at clojure.core$load.doInvoke(core.clj:6034)`

# suggested solution

Something like[ inlining a base64 encoder into the source code](https://github.com/jexler/grengine/commit/659f4fe8a5433339bb6ac591d79a687374687dc7).

One example of the source of the [javax.xml.bind.DatatypeConverter](http://grepcode.com/file/repository.jboss.org/nexus/content/repositories/releases/org.jboss.embedded/thirdparty-all/beta3.SP15/com/sun/xml/bind/DatatypeConverterImpl.java#DatatypeConverterImpl._printBase64Binary%28byte%5B%5D%29) is however a little more elaborate, and makes closer considerations than the code in grengine above on the allocated string buffer size etc.

zentrope:
You can get around this with:

    --add-modules java.xml.bind

or in Lein:

    :jvm-opts ["--add-modules" "java.xml.bind”]

but then you need a conditional build somehow if you intend your project to be backwards compatible given that Java 8 will fail with that particular switch.

claj:
Thanks a lot for that switch. It will work in our particular case.

Are you interested in a patch which either

a) inlines the functionality

[...]

PR #370, opened 2018-03-16
[#356] Java 7-9 base64-encoder support

claj:
macro with clever import logic from @benalbrecht in separate namespace.


claj:
# Manual testing with Java 9
I have Java 8 as default on my mac. I tested with Java 9 using:

`export JAVA_HOME='/Library/Java/JavaVirtualMachines/jdk-9.0.1.jdk/Contents/Home/'`
`lein test`

Use default JVM by:

`unset JAVA_HOME`

ptaoussanis:
Merged manually, thanks :+1:

Issue #345, opened 2017-07-28
Stuck on multiple async calls


Grievoushead:
I have created small example which highlight the problem:

```
(->> (range 0 4)
     (mapv (fn [i]
             (http/get "http://http-kit.org/"
                       (fn [res]
                         (info "first callback")
                         (let [res2 @(http/get "http://http-kit.org/")]
                           (info "second callback ")))))))```
```
 It's stuck on printing the 4s `first callback` msg's.

If I change the range for `0..3` it will work, the sync version also works.

![image](https://user-images.githubusercontent.com/1427311/28731760-d928821e-73dd-11e7-8ed9-bd03a20499eb.png)

Update:
Try with more elements in range if my example worked for you, 10 for example.

afandian:
I wonder if this is connected to a problem I'm having. What does the stacktrace look like?
 https://stackoverflow.com/questions/45619098/threads-parked-with-http-kit

Grievoushead:
@afandian Not sure, but here is how I was able to fix the problem https://stackoverflow.com/questions/45378289/clojure-http-kit-get-request-stuck-on-multiple-async-calls/45390117#45390117

afandian:
Thanks. I don't think I've got a deadlock, but I'll take another look.

Grievoushead:
@afandian do you think that possibility to make another api calls in callback function is something not standard and should be avoided?

afandian:
Looking at the http-kit source, it looks like there's a fixed size threadpool for executing callbacks. I have no particular expertise here, but in my opinion, I would not place http-kit calls in callbacks that execute on that threadpool. Instead, maybe use `core.async` to linearize and decouple the calls.

ptaoussanis:
Just to confirm: yes, http-kit has a limited number of threads for handling callbacks. Please use a threading mechanism of your choice (futures, agents, core.async, etc.) to keep the http-kit threads from being tied up on long-running handlers.

Chouser:
@ptaoussanis Would you be open to a patch that throws an exception when deref'ing an http-kit callback inside an http-kit callback? Doing anything "slow" in a callback could obviously be a performance problem, but deref'ing an http-kit callback makes it very likely to hit a hard deadlock, and might be worth a little bit of code to help users avoid mistakes. What do you think?

ptaoussanis:
@Chouser Hi Chris,

Sure, sounds like a good idea- thank you!

ptaoussanis:
[...]

PR #372, opened 2018-04-17
[#345] Add deadlock-guard for callbacks

Chouser:
To see a real deadlock, call (bad-callback n true) with an n greater than the number of CPUs on your test machine. Relates to ticket #345

### DIFF
--- a/src/org/httpkit/client.clj
+++ b/src/org/httpkit/client.clj
@@ -137,6 +137,18 @@
                                   (format "Invalid event-names: (%s) %s"
                                     (class event-names) (pr-str event-names)))))))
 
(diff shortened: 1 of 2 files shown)


ptaoussanis:
@Chouser This is great Chris, thanks!

One thought: how would you feel about about naming the param `:deadlock-guard?` instead? 

That would avoid the negative (which I find can become confusing), and would introduce a `?` for clarity + consistency with the other params.

ptaoussanis:
Pinging @balajirrao, @kassapo, @giorgio-v out of curiosity (note similarity to recent work you did)

ptaoussanis:
Merging manually now, thanks again!

PR #373, opened 2018-05-03
Fix #306 - use HTTP Connect when working via proxy

### LINKED ISSUE #306: SSL through proxy should use CONNECT method
Most proxy-aware HTTP clients that support SSL use the CONNECT method during initial connection to the proxy, then once SSL handshake has taken place, send the original request (GET, for example) on to the origin server.  The proxy itself does not take part in SSL key exchange or encryption and simply shuffles opaque bytes back and forth.

The behavior I have observed with http-kit is that it simply sends a GET https://origin-server directly to the proxy, rather than preceding it with a CONNECT request.  The following sanitized squid proxy access log shows this behavior:

`TCP_MISS/200 1003 GET https://origin-server/url/path.json? - HIER_DIRECT/192.168.1.1 application/json`

Other http client implementations (other languages, the curl command, browsers, etc) using the CONNECT method generate a log entry like the following:

`TCP_MISS/200 68938 CONNECT origin-server:443 - HIER_DIRECT/192.168.1.1`

For my usage, it's preferable from a security standpoint to use the CONNECT method, if it were supported in the http-kit module.

biancazzurri:


ptaoussanis:
@biancazzurri Hi Roman, this looks really promising- thank you!

Could you give me some idea how extensively this has been tested/used in production, etc.? What's your level of confidence re: readiness to merge?

(Asking since it's not a trivial change, and want to have some idea how extensively you think this should be reviewed before merging).

Cheers :-)

biancazzurri:
Hey

This was not tested in production since we found another solution :) 
I think what should be done is a stress testing, sending requests for a couple of hours to several http and https sites via proxy and directly, and use a profiler to see if some leaks are present. I am ready to do that, but before that I wanted someone else to look at the code.
Before another solution for our problem was found I checked this with my use case - AWS ELB which redirects request to a swarm of proxies, and it was cool.

PR #374, opened 2018-05-11
Add support for binding client requests to a local address

snoe:
On hosts with multiple public ips it can be important to send requests originating from a particular one. This PR adds a bindAddress to HttpClient, which when set, will force the socket to bind to the specified address.

```clojure
(client/get "https://api.ipify.org?format=json"
                {:client
                 (client/make-client {:bind-address (InetSocketAddress. (InetAddress/getByName "192.168.0.165") 0)})}
                (fn [result] (println result)))
```


snoe:
Great! Thank you!

PR #376, opened 2018-05-15
Add duplicated-headers-as-seq? flag to requests

kdavy:
Add an option to requests which forces duplicated keys in the response to be returned as a seq of strings instead of as a single, comma delimited string.  

I am using http-kit to build a reverse proxy for some services.  Some of the requests to those services contain headers with multiple values and http-kit currently concatenates all such multi-headers into a single, comma delimited value.  For example, there can be multiple 'set-cookie' headers in a request and the services are not happy when those headers are merged into one comma separated value.  I can (and I currently do) split those headers up again and turn them into multiple headers, but that seems a bit messier and less effiecient than necessary.  It also requires me to know which headers I need to check and split if necessary and there don't seem to be any real standards around which headers might be duplicated.  So having the capability to optionally leave multiple headers as a seq of headers makes building a reverse proxy easier and more robust.

### DIFF
--- a/src/java/org/httpkit/HttpUtils.java
+++ b/src/java/org/httpkit/HttpUtils.java
@@ -332,7 +332,7 @@ public class HttpUtils {
         System.err.print(str.getBuffer().toString());
     }
 
-    public static void splitAndAddHeader(String sb, Map<String, Object> headers) {
+    public static void splitAndAddHeader(String sb, Map<String, Object> headers, boolean duplicatedHeadersAsSeq) {
         final int length = sb.length();
         int nameStart;
         int nameEnd;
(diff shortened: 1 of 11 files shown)


ptaoussanis:
Hi there,

Thanks for this! Could you please provide some context on motivation? Also, may I ask for the whitespace changes to be removed from the PR?

Much appreciated :-)

kdavy:
@ptaoussanis I have fixed some of the whitespace differences, but some of them seem to me to just be mistakes in the existing code.  For example, in src/java/org/httpkit/server/HttpServer.java there are a couple of lines with an extra space at the end (lines 207 and 209) and in the same file there are a bunch of lines that start with a tab and then a bunch of spaces, all on the same line.  I can change them back if you like, but it seems better to me to fix them.

ptaoussanis:
Hi there, could you please provide some context on motivation? (What is this for, why is it needed, etc.).

Thanks!

ptaoussanis:
Closing due to inactivity.

PR #379, opened 2018-06-17
support byte[] responses

ikitommi:
Why? With libs like [jsonista](https://github.com/metosin/jsonista), the fastest way to serialize Clojure/EDN data is to write it directly into `byte[]`. Aleph and the next version Immutant support this already.

Ported the top3 Clojure Entries in [techempower web benchmarks](https://www.techempower.com/benchmarks/) to Jsonista + `byte[]`: 

screen shot 2018-06-17 at 15 45 07


ikitommi:
Any change of getting this in? Ring will start to support this too: https://github.com/ring-clojure/ring/pull/334

ptaoussanis:
Hi Tommi, thanks for this! Looks good, will include this with the next release 👍

ptaoussanis:
Thanks Tommi, merging no!

Issue #318, opened 2016-12-02
initial websockets message comes before I get a chance to hook on-receive


darwin:
I'm a happy user of http-kit. On server I have a http-kit-powered web-sockets server in Clojure and on browser side I have a web-sockets client in ClojureScript. 

The problem is that my implementation is flaky. My app suffers from occasional and very rare situations when initial `:ready` message sent from client to server gets lost. This usually happens in dev scenario where both client and server run on localhost and both processes are cold.

I have a theory what is going wrong and I'm afraid it is a fundamental issue of http-kit API design.

This is my server code:
```clojure
(defn on-new-client-connection [server request]
  (http/with-channel request channel
    (if-not (http/websocket? channel)
      {:status 200 :body "Please connect with a websocket client!"}
      (let [client (make-client channel)]
        (http/on-receive channel (partial add-job! client :receive server client))
        (http/on-close channel (partial add-job! client :close server client))
        (boot-client! server client)))))
```

Imagine that client sends the initial `:ready` message ASAP after opening connection and this Clojure code is slow for some reason. There is a non-zero time window between `with-channel` and `http/on-receive` calls. If message is fast enough and comes before `http/on-receive` it gets silently lost.

Unfortunately it is pretty hard to reproduce this issue in the above code. But obviously the problem can be amplified by putting something like `(Thread/sleep 1000)` before `http/on-receive`.

After reading http-kit code, I believe there is no way how to hook `http/on-receive` to 100% prevent this scenario, because channel is given to the callback after connection has been initiated. And as far as I'm aware there is no "replay" mechanism, which would replay previously received messages after late `http/on-receive` call. 

Ideas?

For reference full implementation is here:
https://github.com/binaryage/dirac/blob/master/src/lib/dirac/lib/ws_client.cljs
https://github.com/binaryage/dirac/blob/master/src/lib/dirac/lib/ws_server.clj

zhming0:
In my case, I used normal HTTP URI to transfer the initial message. But it means that the initial message has to be short.

marco-nicolini:
I think I'm being hit by this as well in a pet project.
I have the same symptoms: 
- Happens occasionally, especially on cold a server.
- The client (standard browser websocket in js) connects and immediately sends a message (a login message in my case) when the ws connection is opened.

My usage is here https://github.com/marco-nicolini/frothy/blob/master/src/undercover/core.clj

I'm thinking of working around this by changing the application protocol.
[...]

PR #396, opened 2019-03-23
[Experimental] `as-channel` API to address [#318]

ptaoussanis:
This is an attempt to address #318, and an alternative to #391.
In contrast to #391, this approach:

  - Avoids a breaking change by keeping the old (racey) behaviour
    by default.

  - Offers a new API (`as-channel`) with explicit pre/post WebSocket
    handshake handlers.

Another alternative would be to (re)introduce a queue as discussed at
#318 (comment), but would be a fair bit of complexity.

### DIFF
--- a/src/org/httpkit/server.clj
+++ b/src/org/httpkit/server.clj
@@ -91,7 +91,42 @@
       {:local-port (.getPort s)
        :server s})))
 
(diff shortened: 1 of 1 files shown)


huahaiy:
Agree with the analysis, this approach seems to cause the least interruption.

Issue #186, opened 2015-01-26
Server feature request: detect when server is done stopping


tomconnors:
It appears that stopping the http-kit web server is async, since ThreadPoolExecutor.shutdownNow doesn't wait for active tasks to terminate. Using Stuart Sierra's `component` application architecture and a `reload` function doesn't play well with this asynchrony, since stopping then immediately starting a server throws an error due to the address being in use. Is there any way we could be notified when the server is done listening?

venantius:
I'd like to +1 this - in particular, because I think the docs on the subject are misleading.

Quoth the [docs](http://www.http-kit.org/server.html#options):
```clojure
(defn stop-server []
  (when-not (nil? @server)
    ;; graceful shutdown: wait 100ms for existing requests to be finished
    ;; :timeout is optional, when no timeout, stop immediately
    (@server :timeout 100)
    (reset! server nil)))
```

...in point of fact, omitting the `:timeout` key does nothing to force the server to stop, nor does setting a `:timeout` key of 0. In either case, attempting to restart a server in this fashion results in the rather inconvenient
```
BindException Address already in use  java.net.PlainSocketImpl.socketBind (PlainSocketImpl.java:-2)
```

shenfeng:
should  set `SO_REUSEADDR`?

Here is another solution:

`#'` in `#'chartrootm`   make it  possible not needed to stop the server and restart, for reload purpose.

```clj
(defroutes chartrootm
  (GET "/ws" []  chat-handler)
  (files "" {:root "static"})
  (not-found "<p>Page not found.</p>" ))

(defn- wrap-request-logging [handler]
  (fn [{:keys [request-method uri] :as req}]
    (let [resp (handler req)]
      (info (name request-method) (:status resp)
            (if-let [qs (:query-string req)]
              (str uri "?" qs) uri))
      resp)))

(defn -main [& args]
  (run-server (-> #'chartrootm site wrap-request-logging) {:port 9899})
  (info "server started. http://127.0.0.1:9899"))
```

ptaoussanis:
PR welcome that'd provide a shutdown-complete callback (or `[& synchronous?]` shutdown option).

ptaoussanis:
Will be addressed in next release: `server-stop!` call will now return a promise that you can block on if you like.

PR #399, opened 2019-04-08
RFC: Graceful Shutdown

marcomorain:
We've been having some problems dealing with shutdown of http-kit at CircleCI - here is a quick and dirty patch that we are testing presently. I'm opening this PR to get some feedback and share what we are up to - I don't have any expectation of this being merged as is:

## Shutdown more gracefully in the presence of HTTP keep alives.

A two phase approach:

- keep track of the currently kept-alive connections in a set, so that
we can close them when we are shutting down.
- Set a boolean flag true when we are shutting down, so that we can use
it to force isKeepAlive to false when shutting down.

Bonus Features:

Change the error message in RingHandler,.java to be correct when the
ExecutorService is being shutdown. This error message happens when the
server is overloaded _and_ when it is being shutdown.

Add logging to the empty `catch` blocks in HttpServer.java

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject http-kit "2.3.0"
+(defproject circleci/http-kit "2.3.1-SNAPSHOT"
   :author "Feng Shen (@shenfeng)"
   :description "High-performance event-driven HTTP client/server for Clojure"
   :url "http://http-kit.org/"
--- a/src/java/org/httpkit/server/HttpServer.java
+++ b/src/java/org/httpkit/server/HttpServer.java
@@ -10,6 +10,7 @@ import static org.httpkit.server.Frame.CloseFrame.CLOSE_MESG_BIG;
 import static org.httpkit.server.Frame.CloseFrame.CLOSE_NORMAL;
 
 import java.io.IOException;
+import java.io.Closeable;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
 import java.nio.channels.ClosedSelectorException;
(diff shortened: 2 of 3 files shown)


glenjamin:
To add a bit more context, the particular scenario we encountered is an upstream HTTP load balancer (nginx) accessing the application server via a TCP load balancer (kube-proxy).

In this scenario, even when the layer 4 balancer stops accepting new connections, the existing keepalive connections are still active and receive traffic.

The goal of this PR is to cleanly shut down any idle keepalive connections at the same time as we stop accepting new connections.

We have done some canary testing in production, and this appears to work correctly.

ptaoussanis:
Thanks for the clear PR description + context 👍 

In principle sounds good to me. Don't have the opportunity to look too closely at the moment- but let's see if someone else might be able to assist.

Otherwise please keep us updated re: how your testing goes. We can aim to get this in to the next alpha release.

ptaoussanis:
Related to https://github.com/http-kit/http-kit/pull/399

glenjamin:
A small update, we're now running this branch in production, and we're seeing a significant reduction in the number of requests received after the shutdown event - especially when we have longer running requests in-flight.

We're still seeing a non-zero number of received requests, which I intend to delve into more closely when time allows.

ptaoussanis:
Merging manually in current state. @glenjamin Please feel free to open a new PR if you have any further improvements in mind later.

Thanks!

devth:
Is this at all related to https://github.com/http-kit/http-kit/issues/186?

Issue #400, opened 2019-04-18
Option to not decompress GET response body?


chen-factual:
It looks like http-kit automatically sets "accept-encoding" to "gzip, deflate", and it automatically decompresses the data.

If say, I wanted to download a remote gzipped resource, but receive it as gzipped, without the automatic decoding, how do I go about that?

Thanks!

ethpran:
+1!

ptaoussanis:
PR welcome!

MarcoNicolodi:
use `:as :none` options in request map and read the  `org.httpkit.DynamicBytes` bytes using `org.httpkit.ByteInputStream` passing the bytes and the length like

;; body is (:body pedestal-response)
(defn body-bytes->input-stream [^DynamicBytes body]
  (BytesInputStream. (.get body) (.length body)))

now you have a input stream, you can just bypass it or serialize the bytes the way you want.

ptaoussanis:
Thanks Marco!

PR #402, opened 2019-04-27
#400 - add option for no decompressing or coercion whatsoever

ethpran:
":as" key which is used for coercion has a new option called ":none" which will immediately return the body of the response as DynamicBytes which is basically just a byte array.

Got this failure in `rake junit`:
```
There was 1 failure:
1) testDecodeChunkSplits(org.httpkit.client.HttpClientDecoderTest)
java.lang.NullPointerException
```
but it seems to be happening in master as well so I don't think I caused any regression.

### DIFF
--- a/src/java/org/httpkit/client/RespListener.java
+++ b/src/java/org/httpkit/client/RespListener.java
@@ -127,6 +127,10 @@ public class RespListener implements IRespListener {
             return;
         }
         try {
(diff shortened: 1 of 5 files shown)


ptaoussanis:
Hi Ethan, thanks for this! Merged manually. Will try cut a new release next week.

PR #412, opened 2019-09-24
Fix SSLEngine configuration for Java 11

RokLenarcic:
The last fix for Java 11 is faulty:

The person didn't explore the reason behind the bug, instead they introduced a second bug:

https://github.com/http-kit/http-kit/blob/master/src/java/org/httpkit/client/HttpClient.java#L361

The change in newer Java versions is was that `getUseClientMode` will now return `true` if the value wasn't set, so it was never set at the linked location, so SslEngine complained about the mode not being set.

The previous fix changed NOOP configurator as default to one that sets `UseClientMode`. Configurator is called here: 

https://github.com/http-kit/http-kit/blob/master/src/java/org/httpkit/client/HttpClient.java#L364

This is problematic for two reasons:

- this SSLEngineURIConfigurer is supplied as default if caller doesn't supply their own. If caller supplies their own SSLEngineURIConfigurer, they will likely not receive the benefits of this fix since the code in the fix never gets called and they will need to add a call to `setUseClientMode` in their own implementation. Suddenly you have a fix that doesn't work for some people and it's not clear why.

- the bigger problem is that the SSLEngineURIConfigurer is called on each http exec call. Combine with the fact that you also get an exception if you call `setUseClientMode` on an SslEngine that has sent any data, this means that if caller specifies their own SslEngine and reuses it for multiple calls, it will break. See additional unit test.

### DIFF
--- a/src/java/org/httpkit/client/ClientSslEngineFactory.java
+++ b/src/java/org/httpkit/client/ClientSslEngineFactory.java
@@ -8,7 +8,7 @@ import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 
 
-public class SslContextFactory {
+public class ClientSslEngineFactory {
 
     private static final String PROTOCOL = "TLS";
     private static final SSLContext CLIENT_CONTEXT;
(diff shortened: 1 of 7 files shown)


ptaoussanis:
Thanks Rok, merged manually! Will try cut a new release next week.

PR #414, opened 2019-10-02
Update readme with updated tutorial text

doby162:
Bring remote documentation onto github and make updates to namespacing to allow code to run on a lein repl with http-kit specified in the project file.

The documentation is very helpful, but I ran into significant issues trying to run it as someone unfamiliar with the language. (I thought perhaps I was using lein incorrectly!) These small edits make it work for me, and hopefully everyone else!


ptaoussanis:
Hi Michael, thanks for this!

I'd prefer to avoid adding all of this to the README. Perhaps you could add this info to the GitHub Wiki, then we can link to it from the README?

Thanks!

PR #415, opened 2019-10-10
Add error message when no location header is present

### LINKED ISSUE #403: NullPointerException on missing location for 30* responses
Hello

I'm currently dealing with an external api which has a bug where the location header is not returned with the 30* response. This causes a NullPointerException which is returned in the response object.

```Clojure
:error #error {
 :cause nil
 :via
 [{:type java.lang.NullPointerException
   :message nil
   :at [java.net.URI$Parser parse "URI.java" 3042]}]
 :trace
 [[java.net.URI$Parser parse "URI.java" 3042]
  [java.net.URI <init> "URI.java" 588]
  [java.net.URI create "URI.java" 850]
  [java.net.URI resolve "URI.java" 1036]
  [org.httpkit.client$request$reify__15261 onSuccess "client.clj" 228]
  [org.httpkit.client.Handler run "RespListener.java" 42]
  [java.util.concurrent.Executors$RunnableAdapter call "Executors.java" 511]
  [java.util.concurrent.FutureTask run "FutureTask.java" 266]
  [java.util.concurrent.ThreadPoolExecutor runWorker "ThreadPoolExecutor.java" 1142]
  [java.util.concurrent.ThreadPoolExecutor$Worker run "ThreadPoolExecutor.java" 617]
```
This is does not give any insight into what has actually happened the request. I would like to see that error responds with a message saying that there was no location present on the response header.

I can make a PR for this if its worth while.

BadAlgorithm:
fixes #403 

Add in a specific error message when a redirect is returned without a location.

### DIFF
--- a/src/java/org/httpkit/server/AsyncChannel.java
+++ b/src/java/org/httpkit/server/AsyncChannel.java
@@ -290,6 +290,16 @@ public class AsyncChannel {
     static Keyword K_WS_1001 = Keyword.intern("going-away");
     static Keyword K_WS_1002 = Keyword.intern("protocol-error");
     static Keyword K_WS_1003 = Keyword.intern("unsupported");
+    // 1004 is Reserved
+    static Keyword K_WS_1005 = Keyword.intern("no-status-received");
+    static Keyword K_WS_1006 = Keyword.intern("abnormal");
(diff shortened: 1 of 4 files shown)


ptaoussanis:
Thanks Luke! Merged manually. Hoping to try cut a new release next week 👍

PR #419, opened 2019-12-13
Fix/issue 410 ipv6 host header

### LINKED ISSUE #410: "http server loop error, should not happen" with "java.lang.NumberFormatException"
Hi, first, thanks for http-kit. We love it - especially brilliant is how having so few dependencies helps avoid conflicts!

I found the following exception in my logs today, and I've never seen anything like it.

```
Tue Aug 06 17:31:03 UTC 2019 [server-loop] ERROR - http server loop error, should not happen
java.lang.NumberFormatException: For input string: "a9fe]"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.valueOf(Integer.java:766)
	at org.httpkit.server.HttpRequest.setHeaders(HttpRequest.java:79)
	at org.httpkit.server.HttpDecoder.readHeaders(HttpDecoder.java:269)
	at org.httpkit.server.HttpDecoder.decode(HttpDecoder.java:186)
	at org.httpkit.server.HttpServer.decodeHttp(HttpServer.java:154)
	at org.httpkit.server.HttpServer.doRead(HttpServer.java:233)
	at org.httpkit.server.HttpServer.run(HttpServer.java:349)
	at java.lang.Thread.run(Thread.java:748)
```

Some system details, happy to provide any others that may be helpful:
```
$ lein --version
Leiningen 2.9.1 on Java 1.8.0_222 OpenJDK
```

```
http-kit version: 2.3.0
```

Anyone encountered this before? Some googling revealed a similar looking one, but that came with some Key related exception instead of this number parsing one.

Turning everything off and back on again has got me running again, but maybe this report will help uncover a bug.

Thanks again.

blak3mill3r:
special-case the parsing for ipv6, if the host header starts with [

then find the index of ] and if the last : is prior to that,
then do not incorrectly treat the rest of the string as the port

this is an attempt to fix #410

### DIFF
--- a/src/java/org/httpkit/server/AsyncChannel.java
+++ b/src/java/org/httpkit/server/AsyncChannel.java
@@ -290,6 +290,16 @@ public class AsyncChannel {
     static Keyword K_WS_1001 = Keyword.intern("going-away");
     static Keyword K_WS_1002 = Keyword.intern("protocol-error");
     static Keyword K_WS_1003 = Keyword.intern("unsupported");
+    // 1004 is Reserved
+    static Keyword K_WS_1005 = Keyword.intern("no-status-received");
+    static Keyword K_WS_1006 = Keyword.intern("abnormal");
+    static Keyword K_WS_1007 = Keyword.intern("invalid-payload-data");
+    static Keyword K_WS_1008 = Keyword.intern("policy-violation");
+    static Keyword K_WS_1009 = Keyword.intern("message-too-big");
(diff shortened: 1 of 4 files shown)


dchacke:
Any idea when this will be merged?

ptaoussanis:
Merged manually, thanks Blake! Will try cut a new release next week.

Issue #344, opened 2017-07-24
http-kit accepts some invalid SSL certificates


vgeshel:
Example URL: https://78recipes.com/. This site has a certificate with a different hostname, but http-kit accepts it:

```clojure
=> @(org.httpkit.client/request {:url "https://78recipes.com"})
{:status 200 ...}
```

This certificate is rejected by every other HTTPS client I've tried, including other Java/Clojure clients:

```clojure
=> (.openStream (java.net.URL. "https://78recipes.com"))
;; exception
=> (clj-http.client/get "https://78recipes.com")
;; exception: SSLException hostname in certificate didn't match: <78recipes.com> != <*.hostgator.com> OR <*.hostgator.com> OR <hostgator.com>  org.apache.http.conn.ssl.AbstractVerifier.verify (AbstractVerifier.java:238)
```

A similar error was reported in #316. The URLs proposed for testing in that issue still work correctly, e.g., ```@(org.httpkit.client/get "https://wrong.host.badssl.com")``` returns an error. But the URL mentioned at the top does not return an error.

javahippie:
It seems, that "https://wrong.host.badssl.com" is also not working properly, as the message there is `unable to find valid certification path to requested target`. The wrong.host certificate also suffers from an incomplete chain. It seems like the http-kit client is not able to handle this mismatch right now, which is a really bad thing.

jeffh:
I was having this problem with `2.4.0-alpha2` (requesting httpbin) and was looking into a better way to configure stuff for HTTPS. Here’s some prototyping work that seems to work more similarly to `URL.openStream`:

```clj
(defn ssl-configure                                                      
  [^javax.net.ssl.SSLEngine ssl-engine ^URI uri]                          
  (let [^javax.net.ssl.SSLParameters ssl-params (.getSSLParameters ssl-engine)]                                                                    
    (.setUseClientMode ssl-engine true)                                   
    (.setEndpointIdentificationAlgorithm ssl-params "HTTPS")              
    (.setServerNames ssl-params [(javax.net.ssl.SNIHostName. (.getHost uri))])                                                                     
    (.setSSLParameters ssl-engine ssl-params)))                           
                                                                          
(def client (http-kit.client/make-client {:ssl-configurer ssl-configure}))                                                                        
(def ssl-ctx (javax.net.ssl.SSLContext/getDefault))             
                                                                          
(defn engine-for-url [ssl-ctx url]                                       
  (let [addr                                                              
[...]

Issue #388, opened 2018-11-30
OpenJDK 11: "Select exception, should not happen"


cjohansen:
I recently bumped a project from openjdk 8 to 11, and now an outgoing HTTPS request is failing with this exception:

    Fri Nov 30 12:55:02 UTC 2018 [client-loop] ERROR - select exception, should not happen
    java.lang.IllegalStateException: Client/Server mode has not yet been set.
    	at  java.base/sun.security.ssl.SSLEngineImpl.beginHandshake(SSLEngineImpl.java:98)
    	at org.httpkit.client.HttpClient.finishConnect(HttpClient.java:395)
    	at org.httpkit.client.HttpClient.run(HttpClient.java:472)
    	at java.base/java.lang.Thread.run(Thread.java:834)

I built my app with OpenJDK 8, 10, and 11, and the code works fine on 8 and 10, but not 11. 

The request made is roughly:

```cli
(http/get "https://hooks.slack.com/services/..." {:query-params {:token "..."}})
```

http-kit version is 2.3.0

claj:
This could be related to a change introduced even in openjdk 8 u181 where
sun.security.ssl.SSLSessionImpl seems to change its constructor, which leads to some NoSuchMethodError exception.

[https://www.mail-archive.com/openjdk@lists.launchpad.net/msg13114.html](https://www.mail-archive.com/openjdk@lists.launchpad.net/msg13114.html) mentions this stacktrace:

[https://launchpadlibrarian.net/396734572/java-update-8u181-b13-1ubuntu0.18.04.1-svn-failure.txt](https://launchpadlibrarian.net/396734572/java-update-8u181-b13-1ubuntu0.18.04.1-svn-failure.txt)

```
018-10-31 12:25:00.902 FATAL: sun.security.ssl.SSLSessionImpl.<init>(Lsun/security/ssl/ProtocolVersion;Lsun/security/ssl/CipherSuite;Ljava/util/Collection;Lsun/security/ssl/SessionId;Ljava/lang/String;IZ)V
2018-10-31 12:25:00.903 java.lang.NoSuchMethodError: sun.security.ssl.SSLSessionImpl.<init>(Lsun/security/ssl/ProtocolVersion;Lsun/security/ssl/CipherSuite;Ljava/util/Collection;Lsun/security/ssl/SessionId;Ljava/lang/String;IZ)V
[...]

PR #423, opened 2020-01-05
New `default-client-https` behaving similarly to `URL.openStream()`

jimpil:
There is a new delayed Var called `default-client-https`. Using this instead of the standard `default-client` will make your client object behave like `URL.openStream()`, which i think is desirable. Addresses #344  - at the very least on Java11+.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -21,7 +21,7 @@
   ["-Dclojure.compiler.disable-locals-clearing=true"
    "-Xms1g" "-Xmx1g"] ; Testing https require more memory
 
-  :javac-options ["-source" "1.6" "-target" "1.6" "-g"]
+  :javac-options ["-source" "1.7" "-target" "1.7" "-g"]
   :java-source-paths ["src/java"]
   :test-paths ["test"]
   :jar-exclusions [#"^java.*"] ; exclude the java directory in source path
--- a/src/org/httpkit/client.clj
+++ b/src/org/httpkit/client.clj
@@ -73,7 +73,7 @@
      (.setSSLParameters ssl-engine ssl-params))))
 
 (defn- coerce-req
-  [{:keys [url method body insecure? sslengine ssl-context query-params form-params multipart]
+  [{:keys [url method body insecure? sslengine query-params form-params multipart]
     :as req}]
   (let [r (assoc req
                  :url (if query-params
@@ -82,8 +82,6 @@
                           (str url "&" (query-string query-params)))
                         url)
                  :sslengine (or sslengine
-                                ;; allow for SSLContext param
-                                (some-> ssl-context .createSSLEngine)
                                 (when insecure?
                                   (ClientSslEngineFactory/trustAnybody)))
                  :method    (HttpMethod/fromKeyword (or method :get))
@@ -115,6 +113,14 @@
 
 ;;; "Get the default client. Normally, you only need one client per application. You can config parameter per request basic"
 (defonce default-client (delay (HttpClient.)))
+
+(defn make-ssl-engine
+  "Helper for creating instances of SSLEngine
+   from the default, or some custom SSLContext."
+  (^SSLEngine []
+   (make-ssl-engine (SSLContext/getDefault)))
+  (^SSLEngine [^SSLContext ctx]
+   (.createSSLEngine ctx)))
 
 (defonce
   ^{:dynamic true
@@ -182,13 +188,13 @@ an SNI-capable one, e.g.:
     bind-address))
 
 (defonce default-client-https
-  ;; like the `default-client` but with hostname-verification turned on
-  ;; (assumes that requests will carry either an ssl-engine or an ssl-context)
+  ;; similar to the `default-client`, but
+  ;; behaving more like `URL.openStream()`
   (delay (make-client {:ssl-configurer https-configurer})))
 
 (def ^:dynamic ^:private *in-callback* false)
 
-(defn ^:private deadlock-guard [response]
+(defn- deadlock-guard [response]
   (let [e #(Exception. "http-kit client deadlock-guard: refusing to deref a request callback from inside a callback. This feature can be disabled with the request's `:deadlock-guard?` option.")]
     (reify
       clojure.lang.IPending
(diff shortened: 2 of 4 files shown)


project.clj
```diff
@@ -21,7 +21,7 @@
   ["-Dclojure.compiler.disable-locals-clearing=true"
    "-Xms1g" "-Xmx1g"] ; Testing https require more memory
 
-  :javac-options ["-source" "1.6" "-target" "1.6" "-g"]
+  :javac-options ["-source" "1.7" "-target" "1.7" "-g"]
```
jimpil: I was not able to work on this project unless I did this - my IDEA wasn't cooperating otherwise :(.

fpischedda:
Hi, thanks for your great job!
Right now this would introduce another default client, together with the one defined [here](https://github.com/http-kit/http-kit/blob/master/src/org/httpkit/sni_client.clj); what do you think about merging your changes with what has been already done?
I mean, your configurer looks much better compared to what is already in the code base but it would be nice to move it to `org.httpkit.sni-client` namespace or to a newer `org.httpkit.https-client`, what is your opinion about this?

jimpil:
Hi there, 

apologies for the delayed response....

The way I see it, there should be no dedicated namespace for `sni-client`, but rather sni? should be just a flag when creating an ssl client. However, I do understand that this is not really an option at this point... 

I could enhance the `default-client` in sni-client.clj, but it would change the behaviour (one could argue towards the better)...how would you feel about that?

jimpil:
Have a look at the latest commit, and let me know what you think. :)

ptaoussanis:
@jimpil Thanks Dimitrios - are the unit tests passing for you? I appear to be seeing failures in the new tests.

fpischedda:
> Hi there,
> 
> apologies for the delayed response....

Don't worry about the delay :)

> 
> The way I see it, there should be no dedicated namespace for `sni-client`, but rather sni? should be just a flag when creating an ssl client. However, I do understand that this is not really an option at this point...
> 
> I could enhance the `default-client` in sni-client.clj, but it would change the behaviour (one could argue towards the better)...how would you feel about that?

The way is see it is that the flag should not necessarily disappeare and it could stay as an additional option in case you want to specify a client for each request; what I am proposing is to improve what is already in sni-client (and maybe change the name of that namespace) in order to be able to set a default client for the whole application without the need to change each each call to request functions (get, post, etc...); let me know if I need to clarify my point!

jimpil:
Everything passes locally against jdk12, but I just tried jdk8 and I am also seeing:

```
java.lang.IllegalArgumentException: Cannot change mode after SSL traffic has started
 at sun.security.ssl.SSLEngineImpl.setUseClientMode (SSLEngineImpl.java:1990)
``` 
I will investigate...

ptaoussanis:
> @jimpil Everything passes locally against jdk12, but I just tried jdk8 and I am also seeing [...] I will investigate...

👍 Thanks! Current PR otherwise looks good. If you can get the tests passing, I'll merge.

> sni? should be just a flag when creating an ssl client. However, I do understand that this is not really an option at this point...

Motivation for the namespace instead of flag is to retain backwards compatibility with JVMs <= 1.6. Providing a flag would otherwise be nice, but breaks compatibility.

jimpil:
@ptaoussanis  

Ok I fixed the tests, but you may notice that the very first test will return a different kind of exception on JDK8 trough to 11:

```clj
#error{:cause nil,
       :via [{:type java.lang.RuntimeException,
              :message "Delegated task threw Exception/Error",
              :at [sun.security.ssl.Handshaker checkThrown "Handshaker.java" 1519]}
             {:type java.lang.NullPointerException,
              :message nil,
              :at [sun.net.util.IPAddressUtil textToNumericFormatV4 "IPAddressUtil.java" 49]}],
       :trace [[sun.net.util.IPAddressUtil textToNumericFormatV4 "IPAddressUtil.java" 49]
               [sun.net.util.IPAddressUtil isIPv4LiteralAddress "IPAddressUtil.java" 241]
               [sun.security.util.HostnameChecker isIpAddress "HostnameChecker.java" 125]
               [sun.security.util.HostnameChecker match "HostnameChecker.java" 93]
               [sun.security.ssl.X509TrustManagerImpl checkIdentity "X509TrustManagerImpl.java" 455]
               [sun.security.ssl.X509TrustManagerImpl checkIdentity "X509TrustManagerImpl.java" 436]
               [sun.security.ssl.X509TrustManagerImpl checkTrusted "X509TrustManagerImpl.java" 252]
               [sun.security.ssl.X509TrustManagerImpl checkServerTrusted "X509TrustManagerImpl.java" 136]
               [sun.security.ssl.ClientHandshaker serverCertificate "ClientHandshaker.java" 1626]
               [sun.security.ssl.ClientHandshaker processMessage "ClientHandshaker.java" 223]
               [sun.security.ssl.Handshaker processLoop "Handshaker.java" 1037]
               [sun.security.ssl.Handshaker$1 run "Handshaker.java" 970]
               [sun.security.ssl.Handshaker$1 run "Handshaker.java" 967]
               [java.security.AccessController doPrivileged "AccessController.java" -2]
               [sun.security.ssl.Handshaker$DelegatedTask run "Handshaker.java" 1459]
               [org.httpkit.client.HttpsRequest doHandshake "HttpsRequest.java" 89]
               [org.httpkit.client.HttpClient doRead "HttpClient.java" 188]
               [org.httpkit.client.HttpClient run "HttpClient.java" 489]
               [java.lang.Thread run "Thread.java" 748]]}
```
Looking at the stacktrace, it's obvious that the right thing happened, but unfortunately the top level exception is not very informative. [Here](https://bugs.openjdk.java.net/browse/JDK-8211339) is the relevant (fixed) ticket:

jimpil:
Hmm wait there seems to be another issue with JDK12 now :(

ptaoussanis:
@jimpil Thanks. Also think it might be nice to add a 4th URL to the tests that actually succeeds?

jimpil:
Hmm the plot thickens...

1. Java-8/9 works as expected
2. Java-11/12 fails with `java.lang.IllegalStateException: Client/Server mode has not yet been set.` - see #388

jimpil:
Ok so here are some news...

I can make it work consistently with Java-11 and above...Unfortunately, for versions below 11 my 'good' tests (the ones expected to pass) fail during hostname-verification (which can be disabled but still).

In short, I don't think there is a good answer that covers all java versions down to 6 (as the project seems to be trying). Moreover, in light of the new HttpClient in Java 11 and above, I'm not sure how useful the httpkit/client will be going forward. 

I can't spend any more time on this today, but I'm open to feedback...

ps: I'll push one more commit which passes on java11/12

jimpil:
Closing this in favor of #426

PR #443, opened 2020-09-10
Use Java7 compatible method to join string in tests

alekcz:
**Change**
Join strings manually in test. `String.join` was only released in 1.8 so tests fail when running with `--release 7`.
https://docs.oracle.com/javase/8/docs/api/java/lang/String.html 

**Effect**
This allows the test suite to run against Java 7. No impact on actual library. 

**Why this matters**
This will allow us to create Github Actions and thus setup automated tests (and perhaps coverage) for all pushes and pull requests. The Github Actions currently fail because of `--release 7`

### DIFF
--- a/test/java/org/httpkit/server/RingHandlerTest.java
+++ b/test/java/org/httpkit/server/RingHandlerTest.java
@@ -71,7 +71,26 @@ public class RingHandlerTest {
 
     private HttpRequest asHttpRequest(String... requestLines) throws ProtocolException, LineTooLargeException, RequestTooLargeException {
         httpDecoder.reset();
-        String joinedRequest = String.join("\n", requestLines);
+        
+        // String joinedRequest = String.join("\n", requestLines); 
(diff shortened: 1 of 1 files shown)


ptaoussanis:
@alekcz Merged manually, thanks Alexander!

borkdude:
Maybe it makes sense to drop support for Java 7 (eventually) since Clojure 1.10 dropped that too: the minimum requirement for that is Java 1.8 (LTS)

ptaoussanis:
@borkdude That's currently the plan for v2.6 👍 . JVM 6 support is only officially being dropped with v2.5 (release candidate status atm).

There's some institutional users of http-kit that can be quite slow to upgrade.

The major motivation to bump to 1.8+ is to enable client SNI support by default. That is a significant motivation, but not drastically urgent IMO since it can be enabled manually easily enough. As soon as v2.5 final is out (ETA 1-2 weeks), can already cut an early v2.6 release.

PR #445, opened 2020-09-12
Add Github action and status badge

alekcz:
**Change**
Create github action to run tests using Java11

**Effect**
We can now see if pull requests break the build

**Note**
The status badge will show as a missing image until the action runs, there is currently know way to make the badge relative to the repo.


ptaoussanis:
Nice, merged! Thanks!

PR #446, opened 2020-09-15
[WIP] Don't reuse connections when the host changed to be compatible with SNI headers

luizhespanha:
Hello, this is still a WIP and only a POC to collect feedback! If you think this code makes sense I'll try to add some tests.

To give the context of the problem:
We are using an SNI proxy, so for example in AWS Route 53 I have:
my-service-a.xpto.com -> my-sni-proxy.xpto.com
my-service-b.xpto.com -> my-sni-proxy.xpto.com

A client makes a request to my-service-a.xpto.com, then my-sni-proxy receives this request, looks on the SNI headers, and sees that the request if for `my-service-a` and dispatch the request.

The problem that we are facing is:
A client makes a request to my-service-a.xpto.com using httpkit and with keepalive enabled. The request is made and the connection is added to keepalives to be reused, but the addr stored it the IP and PORT of my sni-proxy. Then the client makes a request to my my-service-b.xpto.com in the sequence and as it will also use the proxy, the connection will be reused with the same SNI header, which is my-service-a.xpto.com, and not my-service-b.xpto.com, so the proxy sends the request to the wrong destination.

Change the keepalive to -1 solves this problem for us, but it's tricky for the users to understand and sometimes they even know that the SNI proxy exists.

So I changed the code a little bit to consider the host and the IP and PORT together to decide if we should or not reuse a connection. So, in my example, when the client makes a request to my-service-b.xpto.com, the key will be my-service-b.xpto.com123.123.123.123:443 so we won't reuse the same connection used to invoke my-service-a, but it the client tries to invoke the service-a again, I will be able to reuse the connection.

WDYT? Does it make sense? Do you have some other suggestions to deal with it?

Thanks!

### DIFF
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -220,7 +220,7 @@ public class HttpClient implements Runnable {
                         // Ensure that the key is added to keepalives exactly once on a state transition. There could be cases where decoder reaches
                         // ALL_READ state multiple times.
                         if (oldState != ALL_READ) {
-                            keepalives.offer(new PersistentConn(now + req.cfg.keepAlive, req.host, key));
+                            keepalives.offer(new PersistentConn(now + req.cfg.keepAlive, req.addr, req.host, key));
                         }
                     } else {
                         closeQuietly(key);
@@ -282,8 +282,8 @@ public class HttpClient implements Runnable {
 
 
     public void exec(String url, RequestConfig cfg, SSLEngine engine, IRespListener cb) {
-        URI uri,proxyUri = null;
         String host = null;
+        URI uri,proxyUri = null;
         try {
             uri = new URI(url);
             if (cfg.proxy_url != null) {
@@ -425,7 +425,7 @@ public class HttpClient implements Runnable {
         Request job = pending.peek();
         if (job != null) {
             if (job.cfg.keepAlive > 0) {
-                PersistentConn con = keepalives.remove(job.host);
+                PersistentConn con = keepalives.remove(job.addr.toString() + job.host);
                 if (con != null) { // keep alive
                     SelectionKey key = con.key;
                     if (key.isValid()) {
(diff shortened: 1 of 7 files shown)


ptaoussanis:
@luizhespanha Hi Luiz! Thanks a lot for the clear context, and PR.

My quick first impression is that your approach here sounds reasonable. Would be up for seeing some tests added 👍 

Cheers :-)

src/java/org/httpkit/client/HttpClient.java
```diff
@@ -309,8 +310,10 @@ public void exec(String url, RequestConfig cfg, SSLEngine engine, IRespListener
         try {
             if (proxyUri == null) {
                 addr = addressFinder.findAddress(uri);
+                host = uri.getHost();
             } else {
                 addr = addressFinder.findAddress(proxyUri);
+                host = proxyUri.getHost();
```
andrerocker, line 316: In this case, I think that we need to use something like `proxyUri.getHost() + uri.getHost()`

bsless:
Hi all, what's the status of this PR? 
We have some interest in it as well. If you'd like a few more eyes to look it over we can dedicate some people to it (we: @Appsflyer)

ptaoussanis:
@bsless Hi Ben!

Not sure if Luiz is still available for or interested in this PR (@luizhespanha?) - but next step would be for someone to review and add tests. (Which would be very welcome!)

Cheers :-)

Queen300 (approved):
[H](d)

maorelias:
Hi guys any updates on this? 
I'm also suffering from this issue and I think it's important to fix it because it a counter-intuitive behavior - created the http client with the `sni-client/ssl-configurer` but since keep-alive is enabled, the SNI headers get mixed up from time to time when sending requests to multi-tenant IPs.

ptaoussanis:
@luizhespanha Hi Luiz, it would be great to try and get this work included as part of an upcoming v2.7 release. Do you still have any interest in this? Is there something I or someone else might be able to do to try and assist?

ptaoussanis:
Last call for assistance on this PR, in the hope of possible inclusion for upcoming v2.7.

Things that could help get this merged (in order of low->high effort):

- 👍 that the approach in this PR looks solid (along with @andrerocker's [suggestion](https://github.com/http-kit/http-kit/pull/446#discussion_r488711750)).
- Confirmation if you've successfully used these changes yourself.
- Added unit tests.

Thanks everyone!

ptaoussanis:
PR merged so long for experimental inclusion in v2.7 beta.
Closing for now - but feel free to re-open or add comments if anyone still has any input.

@luizhespanha Thanks for all your work Luiz! 🚀

Issue #224, opened 2015-07-15
Client feature request: support multipart/mixed content-type


wilsonch:
Currently the mime subtype for multipart is fixed to multipart/form-data.

ptaoussanis:
@wilsonch PR welcome, otherwise please feel free to close - thanks! :-)

PR #451, opened 2020-10-28
[Feature] Add multipart/mixed basic support

vtrbtf:
- Related issue: #224

Basic multipart/mixed support by accepting a `multipart-mixed?` flag that forces the `Content-Type` and remove the `Content-Disposition` Header (since it relates to form-data requests only) 

I've added a couple of assertions to check if the request-map is being coerced properly
Tests are passing locally ✔️ 

Any suggestions are welcome :)


ptaoussanis:
Merged into `v2.7` branch, thanks @vtrbtf! Apologies for the slow response.

PR #455, opened 2021-01-14
Fix java.version parsing

aiba:
Recent Azul JDK builds identify themselves as e.g. 16-ea".  This change fixes `parse-java-version` to handle this string.

See https://www.azul.com/downloads/zulu-community/?version=java-16-ea&os=macos&package=jdk


ptaoussanis:
@aiba Merged manually, thanks Aaron! Just pushed `[http-kit "2.5.1"]` to Clojars.

PR #426, opened 2020-02-15
Update SNI client

ptaoussanis:
Based on https://github.com/http-kit/http-kit/pull/423

PR #422, opened 2020-01-03
#393 - workaround and documentation for SNI issue

fpischedda:
This PR contains:
- a change in org.http-kit.client to enable the customization of the default client
- a new namespace org.http-kit.sni-client that incudes a default client with :sni-configurer option
- a bit of documentation in the README.md about the SSLHandshakeException problem and how to address it.

PR #335, opened 2017-05-02
Make way for fixing #187

kumarshantanu:
This PR aims to help fix #187 by allowing custom configuration of the SSLEngine in http-kit client. Taking a cue from what #311 tries to do, this PR creates a hatch in the client implementation so that the optional customization (default: no-op) is invoked on the SSLEngine. This PR does not need Java 8 (though one can use customizations that make use of Java 8 features) and passes all existing tests.

I have tested the PR with SNI-enabled websites https://soundcloud.com and https://osteo15.com using the following customization using Java 8:

```clojure
(ns sni.core
  (:require [org.httpkit.client :as http])
  (:import
    [java.net URI]
    [javax.net.ssl SNIHostName SNIServerName SSLEngine SSLParameters]))

(defn sni-configure
  [^SSLEngine ssl-engine ^URI uri]
  (let [^SSLParameters ssl-params (.getSSLParameters ssl-engine)]
    (.setServerNames ssl-params [(SNIHostName. (.getHost uri))])
    (.setSSLParameters ssl-engine ssl-params)))

(def client (http/make-client {:ssl-configurer sni-configure}))

(comment
  @(http/get "https://soundcloud.com" {:client client}))
```

The above customization might be suitable for a separate repo, say `http-kit.contrib.java8`, requiring Java 8. Alternatively, this customization may be put under a different namespace in the same repo, say `org.httpkit.client.java8`, as long as no other namespace references it in the `ns` block.

Issue #334, opened 2017-04-30
HTTPS support


olessavluk:
Hi,
What is the current support for HTTPS?
I've received some errors when tried to make request:

✅  https://www.google.com.ua/

<br>
⛔  https://clojure.org
<details>
<summary>Received fatal alert: handshake_failure</summary>

```
#error {
 :cause Received fatal alert: handshake_failure
 :via
 [{:type javax.net.ssl.SSLException
   :message Received fatal alert: handshake_failure
   :at [sun.security.ssl.Alerts getSSLException Alerts.java 208]}]
 :trace
 [[sun.security.ssl.Alerts getSSLException Alerts.java 208]
  [sun.security.ssl.SSLEngineImpl fatal SSLEngineImpl.java 1666]
  [sun.security.ssl.SSLEngineImpl fatal SSLEngineImpl.java 1634]
  [sun.security.ssl.SSLEngineImpl recvAlert SSLEngineImpl.java 1800]
  [sun.security.ssl.SSLEngineImpl readRecord SSLEngineImpl.java 1083]
  [sun.security.ssl.SSLEngineImpl readNetRecord SSLEngineImpl.java 907]
  [sun.security.ssl.SSLEngineImpl unwrap SSLEngineImpl.java 781]
  [javax.net.ssl.SSLEngine unwrap SSLEngine.java 624]
  [org.httpkit.client.HttpsRequest doHandshake HttpsRequest.java 98]
  [org.httpkit.client.HttpClient doRead HttpClient.java 139]
  [org.httpkit.client.HttpClient run HttpClient.java 426]
  [java.lang.Thread run Thread.java 745]]}
```
</details>

<br>
⛔  https://letsencrypt.org
(works fine with OpenJDK)
<details>
<summary>unable to find valid certification path to requested target</summary>

```
#error {
 :cause unable to find valid certification path to requested target
 :via
 [{:type javax.net.ssl.SSLHandshakeException
   :message General SSLEngine problem
   :at [sun.security.ssl.Handshaker checkThrown Handshaker.java 1431]}
  {:type javax.net.ssl.SSLHandshakeException
   :message General SSLEngine problem
   :at [sun.security.ssl.Alerts getSSLException Alerts.java 192]}
  {:type sun.security.validator.ValidatorException
   :message PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
[...]

PR #464, opened 2021-04-17
Same `HttpClient` instance should be used when following redirects

sumbach:
Thanks, @ptaoussanis @jimpil @fpischedda and all the other contributors, for the SNI fix/workaround -- upgrading to http-kit 2.5.1 and switching to the new `org.httpkit.sni-client/default-client` has been a breeze.

### Issue

I did encounter an unexpected behavior when overriding `*default-client*` with `(binding ,,,)` -- while the client from the dynamic binding is used for the first/initial request, subsequent requests (i.e., for following redirects) use the root binding instead.

I've included a test case here demonstrating the issue, and would be happy to discuss with you all.

### Workaround

The workaround in my case was to specify `:client` in request options _instead_ of using `*default-client*`:

```
(http/get url {:client (force org.httpkit.sni-client/default-client)})
```

### How to address the issue

* I would recommend adding `:client` to request options when following redirects (so the initial request might fetch a client from `*default-client*` but redirects would _always_ include `:client` in the options map.)
* Alternatively, we could make a binding for `(let [,,, bound-request (bound-fn request) ,,,] ,,,)` and use `bound-request` when following redirects.
* Or we could do both.
* I'm happy to update this PR with any of these changes, but I figured it'd be better to confirm the problem and discuss potential solutions here first.

### Improving the tests?

* I believe these tests demonstrate the issue, but they aren't the clearest or most maintainable code.
* If you have any ideas or feedback, please let me know.

### DIFF
--- a/test/org/httpkit/client_test.clj
+++ b/test/org/httpkit/client_test.clj
@@ -384,7 +384,38 @@
           @(http/get url-1))
         (is (= [{:client-id :var-root :url url-1}
                 {:client-id :var-root :url url-2}]
-               @call-log))))
+               @call-log)))
+      (testing "closed over and used for subsequent requests"
+        (let [wrap-blocking (fn [^org.httpkit.client.HttpClient client p]
+                              (proxy [org.httpkit.client.HttpClient] []
+                                (exec [url cfg sslengine ^org.httpkit.client.IRespListener listener]
+                                  (let [blocking-listener (reify org.httpkit.client.IRespListener
+                                                            (onInitialLineReceived [this version status]
+                                                              (.onInitialLineReceived listener version status))
+                                                            (onHeadersReceived [this headers]
+                                                              (.onHeadersReceived listener headers))
+                                                            (onBodyReceived [this buf len]
+                                                              (.onBodyReceived listener buf len))
+                                                            (onCompleted [this]
+                                                              @p
+                                                              (.onCompleted listener))
+                                                            (onThrowable [this t]
+                                                              @p
+                                                              (.onThrowable listener t)))]
+                                    (.exec client url cfg sslengine blocking-listener)))))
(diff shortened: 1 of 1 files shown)


sumbach:
See also #426 #422 #334 #187 #311 #335

ptaoussanis:
@sumbach Hi Sam, thanks for the really clear + comprehensive bug report - nice work! 👍

> How to address the issue [...]

Hmm, I think I'd lean toward (just) the `bound-fn` approach for two reasons:

1. We may want to offer more binding-level config in future.
2. The user may have their own application-level bindings in place. Could be generally helpful to preserve these on retries.

What do you think, does that seem reasonable?

If you agree, would be happy to see a PR that does this and adds the tests (which look reasonable enough to me).

A couple small requests if we go this route:

- Please reference this issue with a comment in the tests, and where you `let` the `bound-fn`.
- Please add a small mention in the `request` docstring that dynamic bindings will be preserved on retries.

Cheers! :-)

sumbach:
Thanks for the quick reply, @ptaoussanis !

I thought of another potential case (changing the var root binding), so I want to add a test case for this and then get back to you on this conversation. Hoping to pick this up yet this week. Thanks! 😄

ptaoussanis:
@sumbach Hi Sam! Just a heads-up that I'll be prepping soon for an upcoming v2.7 dev release. Would be awesome to try get this PR in if possible. Please let me know if I or anyone can assist with something.

Cheers!

ptaoussanis:
@sumbach Hi Sam! Last call for potentially getting this in to the upcoming http-kit v2.7 release. Would be nice to include, but no pressure if you're busy. Cheers :-)

ptaoussanis:
Fixed on master, closing 👍

Issue #471, opened 2021-08-10
Option to not force Accept-Encoding


MarcoNicolodi:
Hello. 

Im creating a proxy-like service that shouldn't have any opinion about and shouldn't interfere in the request. 

The usage of this service should be transparent and exactly equal to the services, which are a lot and I don't control.

But if the clients do not pass the Accept-Encoding header, http-kit forces it, sending a potentially breaking-change binary response to clients.

I would like to know if you are willing to accept a non-breaking change PR that adds the option to avoid forcing Content-Encoding (that would be false by default).

PR #472, opened 2021-08-13
Option to not force accept encoding

MarcoNicolodi:
Solves #471


MarcoNicolodi:
Hello @ptaoussanis 

Do you have an estimate of when you will be able to review/release this, so we can plan on our side here?

Thanks.

ptaoussanis:
@MarcoNicolodi Hi Marco, thanks for the PR!

I can aim to try review around next weekend (4 September).

Suggesting one change so long:

- Instead of `disable-auto-compression?` (default false), have you considered `auto-compression?` (default true)? Think it might be preferable to avoid negative flags, which can be difficult to parse and lead to double negatives, etc.

Thanks

ptaoussanis:
`[http-kit "2.6.0-alpha1"]` now on Clojars, release info [here](https://github.com/http-kit/http-kit/releases/tag/v2.6.0-alpha1).

Thanks again!

PR #476, opened 2021-09-07
Change buildRequestMap to create a PersistentHashMap directly

bsless:
Approximately 50%-60% faster.

Saves up on allocation via two code paths:
- No need to start by creating a TreeMap (~30% of total CPU)
- PersistentArrayMap.create() will anyway populate a PersistentHashMap
when it gets too big

Also saves the intermediate array allocation and loop of populating the
hash map from it.

### DIFF
--- a/src/java/org/httpkit/server/RingHandler.java
+++ b/src/java/org/httpkit/server/RingHandler.java
@@ -31,6 +31,8 @@ import clojure.lang.IFn;
 import clojure.lang.IPersistentMap;
 import clojure.lang.Keyword;
 import clojure.lang.PersistentArrayMap;
+import clojure.lang.PersistentHashMap;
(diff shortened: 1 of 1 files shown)


bsless:
Also see example flame graph - search `buildRequestMap`
https://github.com/bsless/stress-server/blob/big-rerun/results/httpkit.ring-interceptors.java11.G1GC.svg

borkdude:
PR looks good to me.

bsless:
@ptaoussanis would you mind taking a look? This is a relatively low level MR and I'd appreciate another pair of eyes

ptaoussanis:
@bsless Looks good to me also, thanks Ben!
(And thanks Michiel for reviewing earlier).

PR will be included in next release 👍

Issue #475, opened 2021-09-07
GRAAL native-image error with sun.security.provider.NativePRNG 


askonomm:
Basic info:
JDK 11
Lein 2
http-kit 2.5.3

Trace:

```
Error: No instances of sun.security.provider.NativePRNG are allowed in the image heap as this class should be initialized at image runtime. To see how this object got instantiated use --trace-object-instantiation=sun.security.provider.NativePRNG.
24
Detailed message:
25
Trace: Object was reached by 
26
	reading field java.security.SecureRandom.secureRandomSpi of
27
		constant java.security.SecureRandom@68066dce reached by 
28
	reading field sun.security.ssl.SSLContextImpl.secureRandom of
29
		constant sun.security.ssl.SSLContextImpl$TLSContext@1ea24385 reached by 
30
	reading field javax.net.ssl.SSLContext.contextSpi of
31
		constant javax.net.ssl.SSLContext@9536373 reached by 
32
	scanning method org.httpkit.client.ClientSslEngineFactory.getContextInstance(ClientSslEngineFactory.java:29)
33
Call path from entry point to org.httpkit.client.ClientSslEngineFactory.getContextInstance(): 
34
	at org.httpkit.client.ClientSslEngineFactory.getContextInstance(ClientSslEngineFactory.java:29)
35
	at org.httpkit.client.ClientSslEngineFactory.trustAnybody(ClientSslEngineFactory.java:33)
36
	at org.httpkit.client$coerce_req.invokeStatic(client.clj:68)
37
	at org.httpkit.client$coerce_req.invoke(client.clj:60)
38
	at clojure.instant.proxy$java.lang.ThreadLocal$ff19274a.hashCode(Unknown Source)
39
	at java.util.HashMap.hash(HashMap.java:339)
40
	at java.util.HashMap.get(HashMap.java:552)
41
	at com.oracle.svm.core.jdk.Target_java_nio_charset_Charset.lookup(CharsetSubstitutions.java:74)
42
	at java.nio.charset.Charset.isSupported(Charset.java:503)
43
	at com.oracle.svm.jni.JNIJavaCallWrappers.jniInvoke_ARRAY:Ljava_nio_charset_Charset_2_0002eisSupported_00028Ljava_lang_String_2_00029Z(generated:0)
44
```

I know there was a [issue raised about this](https://github.com/http-kit/http-kit/issues/463) and I've tried all the variations, such as:

```
[...]

PR #477, opened 2021-09-07
Add --initialize-at-run-time to always be present as configuration

askonomm:
This follows the conversation in #475, to add --initialize-at-run-time setting to native-image configuration so that GRAAL builds would not fail and users would not have to add this flag themselves. 

I tested this locally with a project of mine and seemed to work just fine, but would be nice if someone else could also give it a go. Thank you @borkdude for the solution.

### DIFF
--- a/resources/META-INF/native-image/http-kit/http-kit/native-image.properties
+++ b/resources/META-INF/native-image/http-kit/http-kit/native-image.properties
@@ -1,0 +1,1 @@
+Args=--initialize-at-run-time=org.httpkit.client.ClientSslEngineFactory\$SSLHolder


borkdude:
@askonomm Did you test this locally? I don't think the `$` needs to be escaped in the config.

borkdude:
The same escaping is used in the babashka config so it's already been proven to work. I also tested it without the dollar, and this also works. So all good.

@ptaoussanis @bsless I endorse this PR.

borkdude:
To summarize the goal of this PR: the documentation here is a manual action: https://github.com/http-kit/http-kit#class-initialization that the user should apply when compiling with native image. With this PR this setting is applied automatically, along with other manual settings the user may provide.

askonomm:
> @askonomm Did you test this locally? I don't think the `$` needs to be escaped in the config.
> 
> EDIT: oh sorry, I see you _did_ test it locally, but it still surprises me that you needed to escape the dollar.

Ah, sorry about that, I'm not that knowledge of bash scripting et al so I thought I had to add it as-is from README, and it worked for me so didn't even wonder about that. Lesson learned!

Issue #478, opened 2021-09-11
Issue with locking Files served to clients


ikappaki:
There is a side effect with a Files serving optimisation in http-kit, at least on MS-Windows, that can result in preventing hot reloading of js files to the web browser during a cljs development session. The issue is with http-kit locking files that it served of size greater than 20 MB, preventing them to be updated by external processes.

In my case for example, it breaks the shadow-cljs REPL development session when the resulting cljs to js file under development is >= 20 MB. The js app is served to the web browser but http-kit locks the js file. When the developer makes changes in the cljs source code, the code is complied to js and hot reloaded to the web browser. Shadow then also tries to sync the modified js code to the local .js file but because it is locked by http-kit it fails the session, preventing the user from doing any further development.

On windows, the error that shadow-cljs eventually returns is

```
The requested operation cannot be performed on a file 
with a user-mapped section open
```

I assume other cljs hot reloading tools such as figwheel to fail in a similar manner.

My expectation is that http-kit should not put locks on served files that can be updated during a session.

Below is a minimal project that demonstrates that http-kit puts a lock on files that is served that are equal or greater to 20 Mb.

*deps.edn*:
```clojure
{:paths ["src"]
 :deps {org.clojure/clojure {:mvn/version "1.10.3"}
        http-kit/http-kit {:mvn/version "2.5.3"}}}
```

*src/issue/http_kit_mapped_file.cljs*:
```clojure
(ns issue.http-kit-mapped-file
  "Example http-kit program to demonstrate locking of served files when
  their size exceeds 20 Mb."
  (:gen-class)
  (:import [java.io File FileInputStream FileOutputStream])
  (:require [org.httpkit.server :as s]))

(defn ^File file-gen
  "Creates a dummy MB.js File in the current directory of size MB megabytes.

  Returns the File."
  ([Mb]
   (let [file (File. (str Mb ".js" ))
         path (.getAbsolutePath file)]
     (with-open [w (clojure.java.io/writer file)]
       (.write w (str "/** " path " ") )
       (.write w (str (apply str (repeat (* 1024 1024 Mb) "*"))))
       (.write w "/"))
     (println :file-created-at path)
     file)))

(defn -main
  "Starts the server at port 3003."
  [& _]

  (let [port 3003]
    (println :running-server... :port port)

    (s/run-server
     ;; The handler. When REQUEST URI is /NN, with NN being an integer
     ;; < 26, it creates an NN.js File of size NN in the current
[...]

PR #479, opened 2021-09-11
Make MappedByteBufferSize threshold configurable via option

ikappaki:
Hi,

please find in this PR a potential but naive solution for the locking issue described in #478 for making the file size threshold configurable.

The patch introduced a new `:file-mapped-thresh` that lets the developer choose the file size threshold above of which http-kit will choose to memory map Files before serving them to the client.

This will at least allow a developer to prevent file locking of large artefacts during development by having them specify the threshold of which files are mapped into memory.

It has been tested with the  test project submitted in #478, with a slight modification as below fore setting the threshold to 6 MB:

in *src/issue/http_kit_mapped_file.cljs:-main*:
```clojure
     ;; server parameters
     {:port port
      :file-mapped-thresh (* 1024 1024 6)}
```
Thus, requesting a file of 4 Mb at http://localhost/4

and then deleting it works ok without errors:
```
>del 4.js
```

But, requesting a file of 6 Mb at http:///localhost/6

and then deleting it results to an error:
```
>del 6.js
The process cannot access the file because it is being used by another process.
```

i.e. the threshold configured at start up has taken effect.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 > This project uses [Break Versioning](https://github.com/ptaoussanis/encore/blob/master/BREAK-VERSIONING.md) as of **Sep 18, 2020**.
 
+## v2.5.3 (2021 Feb 21)
+
+```clojure
+[http-kit "2.5.3"]
+```
+
+> Non-breaking hotfix release.
+> See [here](https://github.com/ptaoussanis/encore#recommended-steps-after-any-significant-dependency-update) recommended steps when updating any Clojure/Script dependencies.
+
+#### Fixes since `2.5.2`
+
+* [#462 #437] Fix project.clj compiler option to support older JVMs (e.g. Java 8)
+
+
 ## v2.5.2 (2021 Feb 19)
 
 ```clojure
--- a/README.md
+++ b/README.md
@@ -104,5 +104,5 @@ Copyright &copy; 2012-2020 [@shenfeng][] and contributors. Distributed under the
 [API]: http://http-kit.github.io/http-kit/
 [@shenfeng]: https://github.com/shenfeng
 [@ptaoussanis]: https://github.com/ptaoussanis
-[project website]: https://http-kit.github.com
+[project website]: https://http-kit.github.io
 [Break Version]: https://github.com/ptaoussanis/encore/blob/master/BREAK-VERSIONING.md
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 **[CHANGELOG][]** | [API][] | current [Break Version][]:
 
 ```clojure
-[http-kit "2.5.2"]  ; Published by contributors, see CHANGELOG for details
+[http-kit "2.5.3"]  ; Published by contributors, see CHANGELOG for details
 [http-kit "2.1.19"] ; Legacy, published by @shenfeng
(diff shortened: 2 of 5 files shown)


bsless:
Instead of setting a public field, what do you think about these two other possible solutions:
- set the threshold via a JVM property
- set via delayed inner class initialization, similar to [this](https://github.com/http-kit/http-kit/blob/master/src/java/org/httpkit/client/ClientSslEngineFactory.java#L15)

ikappaki:
> Instead of setting a public field, what do you think about these two other possible solutions:

Sure. I think the JVM property option is a better solution, appears to be a better fit for setting low level options.

(My Java writing skills are almost non-existent, so feel free to scrutinize the code)
 
>     * set the threshold via a JVM property

Would something like this do?

```java
public class HttpUtils {

    // Files are memory mapped when above this threshold.
    private static int MAPPED_BUFFER_THRESH_SIZE_BYTES = initMappedThreshold();

    // Returns the threshold in bytes after which files should be
    // memory mapped for efficiency, as calculated from the
    // org.http-kit.mapped-file-threshold jvm property (defaults to 20
    // MB).
    private static int initMappedThreshold()
    {
        return 1024 * 1024
            * Integer.getInteger("org.http-kit.mapped-file-threshold", 20);
    }
    // ...
    public static ByteBuffer readAll(File f) throws IOException {
        int length = (int) f.length();
        if (length >= MAPPED_BUFFER_THRESH_SIZE_BYTES) {
            FileInputStream fs = new FileInputStream(f);
            MappedByteBuffer buffer = fs.getChannel().map(MapMode.READ_ONLY, 0, length);
            fs.close();
            return buffer;
        } else {
            return ByteBuffer.wrap(readContent(f, length));
        }
    }
   //  ...
}

```


>     * set via delayed inner class initialization, similar to [this](https://github.com/http-kit/http-kit/blob/master/src/java/org/httpkit/client/ClientSslEngineFactory.java#L15)

Sorry, can't see how to do this without overcomplicating the code (an inner Properties class with a setter and getter for the field in question, instantiated via an inner static Properties field?). Example implementation sketch is most welcome.

Thanks!

ikappaki:
Hi,

I have made the memory mapped size threshold configurable via a new `org.http-kit.memmap-file-threshold` jvm property as per @bsless  suggestion. Could someone please review the change?

It passes the test

e.g. Configure the test server to only memory map files above 2MB
```
clojure -J-D"org.http-kit.memmap-file-threshold=2" -M -m issue.http-kit-mapped-file
```

Test server creates a 1MB file with http://localhost:3003/1, and user can delete it from the command prompt
```
>del 1.js
```

Test server creates a 2MB file with http://localhost:3003/2, but user can't delete it from the command prompt (file is locked)
```
>del 2.js
The process cannot access the file because it is being used by another process.
```

Thanks,

ikappaki:
> 
> 
> @ikappaki Thanks so much for the report, and for your PR! Looks promising, though I unfortunately don't have the opportunity to review in detail today. Will try take a closer look at this for next release. Otherwise someone else can hopefully help review before then!

Great thanks! I've changed the PR's title for clarity.

ptaoussanis:
This will be included in the next (v2.7) dev branch, thanks!

Issue #401, opened 2019-04-27
When reusing connection, isConnected and timeoutTs are not set properly


ethpran:
I might be wrong about this, so I'm double checking before submitting a fix. When [successfully reusing a connection from the keepalives cache](https://github.com/http-kit/http-kit/blob/master/src/java/org/httpkit/client/HttpClient.java#L423), it doesn't look like the request updates its `isConnected` and `timeoutTs` fields like it does after connecting for the first time in the [setConnected](https://github.com/http-kit/http-kit/blob/master/src/java/org/httpkit/client/Request.java#L41) method. This seems like a bug which could cause unwanted consequences in [clearTimeout](https://github.com/http-kit/http-kit/blob/master/src/java/org/httpkit/client/HttpClient.java#L123) in which we use the wrong timeout as defined in request params and the wrong error message.

xwang1498:
I've proposed PR #481 to fix this.

ptaoussanis:
Manually merging xwang1498's PR now! Thanks for the report @ethpran.

PR #481, opened 2021-09-21
[#401] mark Request as connected when reusing keepalive

xwang1498:
Previously, a new Request that reused an existing keepalive conn was never marked as connected, so it would incorrectly use `connect-timeout` instead of `idle-timeout` when deciding whether to timeout.


xwang1498:
I fixed a bug in my changes and added comments explaining.

PR #484, opened 2021-11-30
Add :start-time to request

niquola:
For monitoring http-kit based applications we need to track time spent in queue and estimate full request time.
This PR adds startTime to request, which is set on `doRead` - not sure this is the right place to do it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ pom.xml*
 /target
 /.idea
 /_site
+.cpcache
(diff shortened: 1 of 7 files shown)


ptaoussanis:
@niquola Thanks for this! Looks reasonable to me, merging manually as part of upcoming v2.7 dev branch. Apologies for the delay!

niquola:
Thank you!

niquola:
The only fix may be use nanoTime instead of millis :)

Issue #469, opened 2021-05-11
SSL error on `doRead`


daniel-quintoandar:
I'm getting `javax.crypto.BadPaddingException` exception which may be caused by [synchronization](https://bugs.openjdk.java.net/browse/JDK-8221218?focusedCommentId=14416634&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14416634). I'm not sure if this bug can be fixed on httpkit or just by updating java version :thinking: 

I'm using
- http-kit **2.5.3**
- java **openjdk version "11.0.11" 2021-04-20**

<details>
<summary>Exception stack</summary>

```
javax.crypto.BadPaddingException: Insufficient buffer remaining for AEAD cipher fragment (2). Needs to be more than or equal to IV size (8) + tag size (16)
  ?, in sun.security.ssl/decrypt
  ?, in sun.security.ssl/decodeInputRecord
  ?, in sun.security.ssl/decode
  ?, in sun.security.ssl/decode
  ?, in sun.security.ssl/decode
  ?, in sun.security.ssl/decode
  ?, in sun.security.ssl/readRecord
  ?, in sun.security.ssl/unwrap
  ?, in sun.security.ssl/unwrap
  ?, in javax.net.ssl/unwrap
  File "HttpsRequest.java", line 35, in org.httpkit.client/unwrapRead
    while ((res = engine.unwrap(peerNetData, peerAppData)).getStatus() == Status.OK) {
  File "HttpClient.java", line 191, in org.httpkit.client/doRead
    read = httpsReq.unwrapRead(buffer);
  File "HttpClient.java", line 494, in org.httpkit.client/run
    doRead(key, now);
  ?, in java.lang/run
```

</details>

miikka:
We're seeing a `javax.net.SSLProtocolException` when using http-kit as a HTTP client. It's a different exception, but seems like it could be a related problem. It happens randomly in production and we do not have a reliable way to reproduce it in isolation.

The linked JDK issues look potentially relevant,. Unfortunately they're fixed in OpenJDK 11.0.12, which is going to be [released in July](https://wiki.openjdk.java.net/display/JDKUpdates/JDK11u#JDK11u-Timelines). Could be worth it to try an Early Access release, though.

* http-kit: 2.5.3 
* Java: Corretto-11.0.11.9.1 (build 11.0.11+9-LTS).

<details>
<summary>Exception stack</summary>
```clojure
{:error #error {
[...]

PR #489, opened 2022-05-24
properly unrecycle the req when kept-alive conn wasn't able to be reused

xwang1498:
- Fixes the SSLProtocolException "Input record too big" errors reported in #469. This is a pretty bad bug. More details [here](https://github.com/http-kit/http-kit/issues/469#issuecomment-1136477850).
- Moves more logic into recycle/unrecycle to make them symmetric.
- Fixes potential bug where req was getting added to requests queue too many times.

### DIFF
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -164,12 +164,7 @@ public class HttpClient implements Runnable {
         keepalives.remove(key);
         // keep-alived connection, remote server close it without sending byte
         if (req.isReuseConn && req.decoder.state == READ_INITIAL) {
-            for (ByteBuffer b : req.request) {
-                b.position(0); // reset for retry
-            }
-            req.isReuseConn = false;
-            req.setConnected(false); // "un"-connect this request if the keepalive conn wasn't able to be reused
-            requests.remove(req); // remove from timeout queue
+            req.unrecycle();
             pending.offer(req); // queue for retry
             selector.wakeup();
             return true; // retry: re-open a connection to server, sent the request again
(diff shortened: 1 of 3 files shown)


src/java/org/httpkit/client/HttpClient.java
```diff
@@ -454,13 +448,13 @@ private void processPending() {
...
                     }
                     ch.configureBlocking(false);
+                    // save key for timeout check
+                    requests.offer(job);
                     boolean connected = ch.connect(job.addr);
                     job.setConnected(connected);
                     numConnections++;
                     // if connection is established immediatelly, should wait for write. Fix #98
                     job.key = ch.register(selector, connected ? OP_WRITE : OP_CONNECT, job);
-                    // save key for timeout check
```
xwang1498, line 462: `job.setConnected()` may add to the `requests` queue, so we don't want to add a duplicate afterwards.

ptaoussanis:
Merged manually and included in https://github.com/http-kit/http-kit/releases/tag/v2.6.0-RC1

@xwang1498 Thanks again! Do you think this is something that a test could be added for, it would that be difficult?

xwang1498:
> @xwang1498 Thanks again! Do you think this is something that a test could be added for, it would that be difficult?

@ptaoussanis Glad I could make a good contribution to http-kit. I'm happy to try to add a test for it, but I don't think I'll have time to get to it soon. The good news is this bug is very obvious to a heavy user of http-kit if it ever returns.

xwang1498:
It's entirely up to you, but I'd recommend closing this issue. Don't think it's worth the complexity and hassle of adding tests for this. Would rather people work on higher-value contributions.

PR #494, opened 2022-07-12
Edited formatting of README

ImOwly:
Removed bullet point from the features section, since the bold words at the beginning of the points already do distinguish each feature individually so it does not need to be in bullet point to further imply that, removing the bullet point makes it feel less cluttered and easier to read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,17 +21,17 @@ See the (unmaintained, outdated) [project website][] for original documentation,
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


ptaoussanis:
Hi there, thanks for this! TBH I'm not sure I agree that this change is an improvement. I guess it might be subjective.
I'm inclined therefore to leave as-is unless someone else chimes in.

In any case, appreciate the time+effort! Cheers :-)

PR #495, opened 2022-07-22
CHANGED - modify http server loop error

tweeks-reify:
Wanted to clarify to ensure this error is seen as actionable

### DIFF
--- a/src/java/org/httpkit/server/HttpServer.java
+++ b/src/java/org/httpkit/server/HttpServer.java
@@ -387,7 +387,7 @@ public class HttpServer implements Runnable {
                 // jvm can catch any exception, including OOM
             } catch (Throwable e) { // catch any exception(including OOM), print it
                 status.set(Status.STOPPED);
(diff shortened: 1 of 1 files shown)


ptaoussanis:
@tweeks-reify Hi there, thanks for this!

While I agree that "should not happen" is too vague, I wonder if we could maybe do even better than "investigate this error"?

What kind of error were you running into here exactly?

tweeks-reify:
Thanks @ptaoussanis.  Our issue was caused by running out of Direct Buffer Memory in the JVM (which was captured in the stack trace).  Open to suggestions but maybe even `see stracktrace:` or something like that to indicate there are more details?

ptaoussanis:
@tweeks-reify That sounds reasonable to me 👍 
So "http server loop error, see stack trace for details"?

ptaoussanis:
👍 Will merge for v2.7 next time I'm doing batched work on http-kit - thanks!

ptaoussanis:
Merged manually, closing 👍

Issue #498, opened 2022-11-01
Content-Length header set on Websocket upgrade request 


restenb:
HTTP requests with empty body will have `Content-Length 0` set by default by the HttpEncode class, whenever `Transfer-Encoding` is not present. See the following snippet from HttpUtils.java, line 454: 
            
```
if (!CHUNKED.equals(headers.get("Transfer-Encoding"))) {
   if (bodyBuffer != null) {
      // trust the computed length
      headers.putOrReplace(CONTENT_LENGTH, Integer.toString(bodyBuffer.remaining()));
   } else {
      headers.putOrReplace(CONTENT_LENGTH, "0");
   }
}

```

This will include, among other things, the WebSocket upgrade response. Setting `Content-Length` on this was causing problems with the HTTP proxy on Azure App Service, which terminated the connection instead of relaying the response. 

While Microsoft might fix the above, we also noticed this from the standard:
https://www.rfc-editor.org/rfc/rfc7230#section-3.3.2

> A server MUST NOT send a Content-Length header field in any response
> with a status code of 1xx (Informational) or 204 (No Content).
> 

https://www.rfc-editor.org/rfc/rfc7230#section-3.3.2
>    1.  Any response to a HEAD request and any response with a 1xx
>        (Informational), 204 (No Content), or 304 (Not Modified) status
>        code is always terminated by the first empty line after the
>        header fields, regardless of the header fields present in the
>        message, and thus cannot contain a message body.

So at the very least `Content-Length` seems superfluous in these cases, even though the body size _is_ 0 bytes, and it may cause issues with proxies as we have seen with Azure App Service. 

We also tested two other server implementations; spring websockets and jetty 10. Neither sets `Content-Length` on the upgrade request. So we suggest a change here that takes at least the Websocket upgrade request headers into account before setting `Content-Length`. We have a working prototype of this that we can submit as a PR.

ptaoussanis:
@restenb Hi Roger, thanks for the clear and detailed report 👍 
Would love to see a PR, thanks.

Planning to cut a major new http-kit release in the next few weeks - so would be a great time to get this addressed.

davidhjp01:
Created a pull request at https://github.com/http-kit/http-kit/pull/499

ptaoussanis:
Merging your PR now, thanks! 👍

PR #499, opened 2022-11-01
Websocket fix

davidhjp01:
A fix to the Websocket terminating due to having `Content-Length: 0` in the header, when the application is deployed to Azure App Service. Detailed description can be found at: https://github.com/http-kit/http-kit/issues/498

According to RFC 7230:
```
   A server MUST NOT send a Content-Length header field in any response
   with a status code of 1xx (Informational) or 204 (No Content).  ...
```

In this fix, we check status code passed to `org.httpkit.HttpEncode` and skips adding the content-length header if the status is either `1xx` or `204`.

### DIFF
--- a/src/java/org/httpkit/HttpUtils.java
+++ b/src/java/org/httpkit/HttpUtils.java
@@ -456,7 +456,7 @@ public class HttpUtils {
                     // trust the computed length
                     headers.putOrReplace(
                         CONTENT_LENGTH,
-                        (bodyBuffer != null) ? Integer.toString(bodyBuffer.remaining()) : "0");
+                        (bodyBuffer != null)) Integer.toString(bodyBuffer.remaining()) : "0");
                 }
             }
         } catch (IOException e) {
--- a/src/java/org/httpkit/HttpUtils.java
+++ b/src/java/org/httpkit/HttpUtils.java
@@ -452,11 +452,11 @@ public class HttpUtils {
             bodyBuffer = bodyBuffer(body);
             // only write length if not chunked
             if (!CHUNKED.equals(headers.get("Transfer-Encoding"))) {
-                if (bodyBuffer != null) {
+                if ((status / 100) != 1 && status != 204) {
                     // trust the computed length
-                    headers.putOrReplace(CONTENT_LENGTH, Integer.toString(bodyBuffer.remaining()));
-                } else if ((status / 100) != 1 && status != 204) {
-                    headers.putOrReplace(CONTENT_LENGTH, "0");
+                    headers.putOrReplace(
+                        CONTENT_LENGTH,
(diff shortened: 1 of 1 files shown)


src/java/org/httpkit/HttpUtils.java
```diff
@@ -455,7 +455,7 @@ public static ByteBuffer[] HttpEncode(int status, HeaderMap headers, Object body
                 if (bodyBuffer != null) {
                     // trust the computed length
                     headers.putOrReplace(CONTENT_LENGTH, Integer.toString(bodyBuffer.remaining()));
-                } else {
+                } else if ((status / 100) != 1 && status != 204) {
```
jacobemcken, line 458: Disclaimer: I am no Java expert nor very proficient in the HTTP protocols and servers or the http-kit code in general

From the RFC:
> A server MUST NOT send a Content-Length header field in any response 
>  with a status code of 1xx (Informational) or 204 (No Content).

If for some reason `bodyBuffer` would be non-`null` the code would still send content length for status codes 1xx and 204. :shrug: 

Consider the following to make the RFC implementation more explicit:
```java
if ((status / 100) != 1 && status != 204) {
    // trust the computed length
    headers.putOrReplace(
        CONTENT_LENGTH,
        (bodyBuffer != null)) Integer.toString(bodyBuffer.remaining()) : "0"
    );
}
```

really brilliant with the `(status / 100) != 1` btw.

davidhjp01: Thanks, comitted a new change based on your suggestion. 

Unfortunately, I cannot test the new commits on my end because I am currently away from my workstation (although I don't see any problem with this change). If anyone can test it would be great (@restenb?).

restenb: I don't think the RFC mentions the case of a 101 response having a body, but it would be unnecessary and perhaps unexpected to receive a body in a 101 response, no? If the request has a body, that introduces the need to think about setting `CONTENT_LENGTH` again.

I think the original implementation was probably fine, and servers would be incorrect in sending a 101 response with a body.

ptaoussanis: @davidhjp01 Hi David, thanks for your PR! I'm prepping for the upcoming http-kit v2.7 release and would love to get this included.

What's the current status of the PR, after @restenb's [last comment](https://github.com/http-kit/http-kit/pull/499#discussion_r1045516042)?

Thanks!

davidhjp01: I changed it back to the original version. We can also omit `bodyBuffer` if the passed status is 1xx or 204 but since this is an internal API I guess this is not much an issue (i.e. knowing that we should not pass non-null `body` with 1xx or 204).

ptaoussanis:
@davidhjp01  Thanks David, much appreciated! Merging now 👍

Issue #501, opened 2022-12-13
HTTP Client suggestion for cookies and multiple headers


g23:
Hi, I recently wanted to switch to `http-kit` over `clj-http` since it seems to have better performance, but I noticed handling cookies is tricky. The resulting `:set-cookie` header has all of the cookies concatenated with `,` but for example in a cookie's Expires field there can also be `,`s so it makes it tricky to parse.

My suggestion is to instead concatenate multiple headers with `\n` instead so then the end user can easily split on `\n` to get the multiple values.

I was going to try to fork and make a PR but when I run `lein test` locally on the unchanged code I get 8 test fails. So I'm not sure about my setup.

The change I'm suggesting is `java/org/httpkit/HttpUtils.java` in the `splitAndAddHeader` method, line 378 replace the `","` for `"\n"`

This then results in 3 new test failures that can be fixed by changing `test/org/httpkit/client_test.clj` tests: `test-header-multiple-values` and `test-headers-stringified` to use `\n` instead of `,` on lines 457, 458, and 465.

ptaoussanis:
Hi Greg! PR welcome, I'll look at getting any issues with the tests sorted out when I'm working on the next release 👍

ptaoussanis:
@g23 Your PR has been merged, thanks Greg! 👍

PR #502, opened 2022-12-13
join multiple headers with \n rather than ,

g23:
To make it easier to parse multi-headers like `:set-cookie` see issue #501

### DIFF
--- a/src/java/org/httpkit/HttpUtils.java
+++ b/src/java/org/httpkit/HttpUtils.java
@@ -375,7 +375,7 @@ public class HttpUtils {
             Object v = headers.get(key);
             if (v != null) {
                 // https://github.com/http-kit/http-kit/issues/108
-                value = v.toString() + "," + value;
+                value = v.toString() + "\n" + value;
             }
             headers.put(key, value);
         }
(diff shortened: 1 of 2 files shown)


ptaoussanis:
@g23 PR merged, thank you Greg! 👍

ptaoussanis:
@g23 Hi Greg! This topic has come up again because of #547, so I'd like to understand the original motivation a little better.

At #501 you wrote:

> Hi, I recently wanted to switch to http-kit over clj-http since it seems to have better performance, but I noticed handling cookies is tricky. The resulting :set-cookie header has all of the cookies concatenated with , but for example in a cookie's Expires field there can also be ,s so it makes it tricky to parse.

To make sure we're on the same page, could you please give a code example of the difficulty you were originally running into?

In what way was handling easier in clj-http?

Where did the idea to use "\n" as a separator come from - is that approach being used elsewhere (clj-http, etc.)?

Thanks!

ptaoussanis:
Never mind, my previous message can be ignored - apologies for the noise. Full details of my initial confusion at #547 :-)

PR #503, opened 2023-02-12
properly close sockets when calling stop() on client

benbramley:
When the stop() method is called on the client this stops the client-loop thread and closes the selector however this is not enough to actually close the sockets being used by the client. There are usually sockets left open when using keep-alive.

This is not usually an issue when using the http-kit client on a standalone process as the socket gets killed by the OS when the jvm is shutdown, however in a J2EE/Servlet container where an application can be restarted without stopping the JVM this can leave behind an open connection. This connection is usually eventually cleaned up by a full GC but this PR is a simple change which just cleanly closes the sockets that the http-client is using when the stop() method is called.

### DIFF
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -503,6 +503,9 @@ public class HttpClient implements Runnable {
     public void stop() throws IOException {
         running = false;
         if (selector != null) {
+            for (SelectionKey selectionKey : selector.keys()) {
+                selectionKey.channel().close();
+            }
             selector.close();
         }
     }


ptaoussanis:
@benbramley Hi Ben, thanks for this! Will take a proper look at this when I'm next on batched http-kit work (soon, I'm overdue).

A question in the meantime: do you believe there's any plausible reason why anyone might want your proposed behaviour to be optional? Or stated differently: any reason why the current behaviour may have been intentional? Or are you confident that the current behaviour was simply an oversight?

(Disclaimer: I haven't looked at this closely enough to have any opinion, just think it'd be good to clearly document our thinking on this before making a change).

Cheers!

benbramley:
@ptaoussanis Hey Peter, from what I understand the selector is tied to the HttpClient object so when stop() is called on that object the sockets that are left open would only be the ones owned by that HttpClient object. 
So I don't think there is any reason to make this optional as those sockets are then left orphaned anyway. But when you do get around to looking at this see if you come to the same conclusion, happy to add a config option to the pull request.

Thanks!

Ben

ptaoussanis:
@benbramley Merged, thanks Ben! 👍

PR #504, opened 2023-02-20
Add clj-kondo support to with-channel

figurantpp:
The macro `with-channel` does not play very well with clj-kondo by default. By adding a hook and a config edn file it is possible to get better linting with it.

It currently reports the channel symbol as unresolved.
image

But it is possible to treat it as a proper declaration with the provided changes:
image

### DIFF
--- a/resources/clj-kondo.exports/http-kit/http-kit/config.edn
+++ b/resources/clj-kondo.exports/http-kit/http-kit/config.edn
@@ -1,0 +1,3 @@
+
+{:hooks
+ {:analyze-call {org.httpkit.server/with-channel httpkit.with-channel/with-channel}}}
--- a/resources/clj-kondo.exports/http-kit/http-kit/httpkit/with_channel.clj
+++ b/resources/clj-kondo.exports/http-kit/http-kit/httpkit/with_channel.clj
@@ -1,0 +1,31 @@
+
+(ns httpkit.with-channel
+  (:require [clj-kondo.hooks-api :as api])
+  )
+
+(defn with-channel [{node :node}]
+
+  (let [[request channel & body] (rest (:children node))]
+
+    (when-not (and request channel)
+      (throw (ex-info "No request or channel provided" {})))
+
+    (when-not (api/token-node? channel)
+      (throw (ex-info "Missing channel argument" {})))
+
+
+    (let [new-node
+          (api/list-node
+           (list*
+            (api/token-node 'let)
+            (api/vector-node [channel (api/vector-node [])])
+            request
+            body
+            ))
+          ]
+
+      {:node new-node}
+      )
+    )
+
+  )


ptaoussanis:
@figurantpp Hi there, thanks for this! `with-channel` has been [deprecated in favour of the `as-channel` function](http://http-kit.github.io/http-kit/org.httpkit.server.html#var-with-channel) since http-kit [version 2.4.0](https://github.com/http-kit/http-kit/releases/tag/v2.4.0) several years back.

A future version of http-kit is likely to remove support for `with-channel` altogether.

And since migrating from the old to new API is straightforward, I'm hesitant to add extra code related to the deprecated API.

What do you think?

ptaoussanis:
@borkdude Hi Michiel, thanks for the input.

Yes, my two cents would be that it's better to encourage folks to update.

Motivation:
- `with-channel` has known race conditions
- Migration should be easy in all cases
- There's been a reasonable amount of time to update (since ~Aug 2020)
- `with-channel` will likely be removed soon

One note: I can see that `with-channel` doesn't currently have `^:deprecated` metadata, so it's possible some folks missed the change. I'll add that in the next release.

@kumarshantanu What do you think? I'm happy to hear counter-arguments.

One option would be to merge this code and add the `^:deprecated` meta for the forthcoming `v2.7`, then aim to remove in `v2.8` later in the year.

borkdude:
@ptaoussanis Adding `:deprecated` metadata will be good since clj-kondo will also start to warn about it. 

In addition to your proposal, I would propose a small campaign to make issues in these projects to migrate:

https://grep.app/search?q=with-channel&filter[lang][0]=Clojure

It would help to have an example how to migrate.

kumarshantanu:
@ptaoussanis @borkdude Noted https://github.com/kumarshantanu/ring-sse-middleware/issues/4 - I will look into it.

ptaoussanis:
Thanks Shantanu, much appreciated 👍

So then to summarise the current plan:

- Forthcoming http-kit v2.7 will include:
  - The necessary `^:deprecated` metadata. Will also look at maybe emitting a warning on use.
  - A reminder of the deprecation in the release notes, along with an example migration.
  - Temporary `clj-kondo` support for the deprecated API (i.e. merge this PR).

- The deprecated API will be retained for now. We can re-evaluate removal later based on:
  - How many folks are still using the old API by then.
  - How they're likely to be impacted by the potential race condition in the deprecated API. I'd like to start by taking a closer look at consumers from Michiel's [link]([https://grep.app/search?q=with-channel&filter[lang][0]=Clojure](https://grep.app/search?q=with-channel&filter%5Blang%5D%5B0%5D=Clojure)) when I'm next on batched http-kit work (ETA: within the next month or so).

- Assistance welcome if anyone feels like opening issues with projects still using the old API.

As always, other suggestions also welcome.

borkdude:
Another suggestion: the clj-kondo hook could emit a warning that the function should be replaced by `as-channel` instead. cc @figurantpp

ptaoussanis:
@borkdude 

Thanks for the explanation 👍 

Have you considered the possibility of just printing the docstring of deprecated vars (which I would expect should usually provide details regarding deprecation), and/or encouraging the use of some other standard metadata key for this kind of info (something like `^:prefer` or `^:deprecated-details`, etc.) if you want to avoid irrelevant docstring info?

For example, the docstring for `with-channel` already reads:
```clojure
"DEPRECATED: this macro has potential race conditions, Ref. #318.
  Prefer `as-channel` instead."
```

In general, I'd hope/expect someone to check a deprecated var's docstring to better understand why it's deprecated, implications, suggested migrations, etc. I wonder if it might be better to encourage users to develop that habit (and/or support some standard meta key for details) rather than manually duplicate these kinds of details in hooks?

To clarify: I have zero objection to adding the info to the hook here, have just noticed more users of my own libs using kondo - and I'd like to support them. Adding manual hooks for every deprecation won't scale in my case, so I'm wondering what alternatives have been considered.

Cheers :-)

borkdude:
> Adding manual hooks for every deprecation won't scale 

That's not what I am advocating for either, it's just that we already have this hook in this PR so it would be easy to just add that line. But I'd be satisfied with just the `^:deprecated` metadata. AFAIK there isn't a standard to describe deprecation reasons but personally I'd read the docstring when I find a surprising deprecation warning :)

borkdude:
I think we could even put this information just in the `^{:deprecated "2.0.0: use as-channel instead"}`.

Screenshot 2023-02-20 at 15 08 01

figurantpp:
About the deprecation notice, I was actually not aware that with-channel was deprecated when I wrote the PR. Thanks for the note, I'm now updating the usages in my earlier projects. And honestly, I think it is sufficient to leave a `^:deprecated` message that mentions the `as-channel` alternative on `with-channel`.

borkdude:
One more thing to add: when looking how to use functionality in a library/project I usually first go to the README but if I can't find any references, I look at the tests. The `ws_test.clj` namespace exclusively has examples of `with-channel` and doesn't use `as-channel` at all at this moment, which may lead people to still use `with-channel` instead.

ptaoussanis:
PR merged, thanks @figurantpp!

PR #505, opened 2023-02-27
Prevent duplicate headers in HeaderMap

kipz:
Some tests in `client_proxy_test.clj` are failing for me:

https://github.com/http-kit/http-kit/blob/d12b1667a62785c5bcc635462958a5f85f7db303/test/org/httpkit/client_proxy_test.clj#L122

I _think_ this change https://github.com/http-kit/http-kit/commit/3fbeda1b31d90efb0ff4c208a61328ee0cf5b305 made this appear when `content-length` and `Content-Length` headers are both present. I think we were getting away with it before we started using commas.


ptaoussanis:
@kipz Hi James, thanks for this! Will take a closer look soon, for inclusion in the upcoming v2.7 release. Much appreciated.

ptaoussanis:
Merged manually, thanks! 👍

Issue #461, opened 2021-02-11
Unix domain socket support


tom-adsfund:
Unix domain sockets can be connected to from NGINX. From [this example on Stackoverflow](https://stackoverflow.com/questions/58455529/how-do-i-connect-to-a-unix-domain-socket-running-an-http-server-using-netty) Netty allows using Unix domain sockets.

This could provide substantial performance improvements (from no TCP management).

borkdude:
Maybe relevant, not sure: https://github.com/into-docker/unixsocket-http

tom-adsfund:
@borkdude That would be a client-side. Although Netty can handle it, as in the example. (Big fan of bb btw!)

borkdude:
@tom-adsfund http-kit isn't based on Netty. Aleph is though, which is a different project. And thanks! :)

ptaoussanis:
@tom-adsfund Could you please clarify what the current status is here? As Michiel mentioned, http-kit does not use Netty - so I believe that support for domain sockets would need to be implemented some other way.

But before pursuing that idea further- we'd need to start by clarifying what the particular problem/s are that we'd be hoping to address with domain sockets.

Is there some reason to believe that performance is bottlenecked on the TCP connection between nginx and http-kit server? In my experience that would be highly unusual, and could indicate an nginx configuration issue.

tom-adsfund:
@ptaoussanis It's really just based on the assumption that the OS would be able to manage things better using domain sockets and NGINX and http-kit would be able to multiplex over the domain socket (avoiding many file descriptors), rather than involving the whole TCP stack. Whether that applies in any specific case and implementation pairings etc, I don't know.

The exploration of whether this would work and how it compares I guessed would be under the scope of this library, based on its prior achievements in performance.

ptaoussanis:
Okay- thanks for clarifying 👍 

So http-kit tends to strongly prioritise simplicity, which is both a source of its performance and (I believe) valuable for its own sake.

When it comes to performance, the key questions are:

1. How much of a benefit might some change provide (to what metrics, and under what conditions)?
2. What's the cost? (E.g. in terms of added complexity or other tradeoffs).

So in practice we'd need to start with at least a back-of-napkin estimate for [1] to know if it's worth investigating further.

At least for request->response time, my own intuition says that the nginx<->http-kit TCP comms is very unlikely to be a practical bottleneck since for Clojure applications it's almost always the application handlers that dominate.

[...]

PR #506, opened 2023-02-27
Add support for Unix Domain Sockets

kipz:
This draft hacks in support for Unix Domain Sockets in client and server to initiate a discussion around the approach, especially around whether it's feasible to have a JDK17 only version...

Any ideas or feedback most welcome.

See #461 for a discussion around why people might like UDS support

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,37 @@ This default may be changed in a future breaking release. In the meantime, manua
     <...>)
 ```
 
+## Using Unix Domain Sockets in client and server
+
+Java native Unix Domain Sockets are available since Java 17, so to maintain backwards compatibility with JVMs prior this,
+resolving SocketAddress's and SocketChannels is pluggable in both client and server.
+
+### client
+
+To talk HTTP via a Unix Domain Socket at `/tmp/test.sock`, create your own implementations of `:address-finder` and `:channel-factory` in a customised client:
+
+```clojure
+(let [client (client/make-client {:address-finder (fn [uri]
+                                                    (UnixDomainSocketAddress/of "/tmp/test.sock"))
+                                  :channel-factory (fn [address]
+                                                      (SocketChannel/open StandardProtocolFamily/UNIX))})]
+   (client/get "http://foobar"))
+```
+
+### server
+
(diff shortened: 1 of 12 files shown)


borkdude:
Very interesting. Another interesting experiment might be to leverage the JDK19 virtual threads :)

kipz:
Yeah, it's interesting that it's mostly a search and replace of `InetSocketAddress` with `SocketAddress` :)

kipz:
@ptaoussanis et al - I'm thinking of updating this PR add support for custom address/socket handling when creating clients/servers, and then create a separate mini-library for JDK17+ for unix socket support, but I'm wondering whether or not it's crucial to maintain the current Java APIs? Are they considered part of the public API or not? It's not the end of the world if they are, it's just that it'll end up much cleaner if I can make some changes. WDYT?

ptaoussanis:
@kipz Hi James,

> I'm wondering whether or not it's crucial to maintain the current Java APIs? Are they considered part of the public API or not?

I can't say off-hand, sorry. It might depend on what specific parts of the API you have in mind. Do the parts in question have any docs/comments to indicate whether they're public?

If there's no guidance, a break could well be warranted - depending approximately on:

- Does a web search turn up any obvious users?
- How user-facing is the particular API? The more obscure, the less likely that someone's depending on it.
- If we believe it's reasonable that someone might be depending on the particular API - how easy would it be for them to migrate, and how easily could we document the necessary migration?
- How significant is the benefit of breaking?

**Re: the PR in general**, the clearer the following info can be for me and other reviewers, the easier it'll make reviewing:

1. What's the intended benefit of the PR? Can we be precise? If performance is a significant motivator, can the effect of the PR be quantified?

2. Are there any _known disadvantages_ of the PR? E.g. - are pre-existing users potentially impacted? If so, how? Will anyone be broken? How badly and under what conditions? Is there any particular new maintenance burden we might be incurring?

3. What's our level of confidence that the above disadvantages are exhaustive, i.e. that we're not missing anything important? Are there any particular areas of possible concern that might warrant further input or digging?

Really appreciate all your effort on this 👍
Cheers :-)

kipz:
@ptaoussanis that's my answer right there! :) I can extend what is there without breaking anything!

kipz:
I'll re-open with details of the new approach.

PR #510, opened 2023-03-02
Add support for customizing address and channel providers

kipz:
In support of #461 

* Java/Clojure API's remain unchanged (purely additive)
* listen/bind address and channel providers are customizable in client and server (following current pattern)
* Added an example of using Unix Domain Sockets with JDK17+ to docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,37 @@ This default may be changed in a future breaking release. In the meantime, manua
   (binding [org.httpkit.client/*default-client* sni-client/default-client]
     <...>)
 ```
-
 See `org.httpkit.client/*default-client*` docstring for more details.
+
+## Using Unix Domain Sockets in client and server
+
+Java native Unix Domain Sockets are available since Java 17, so to maintain backwards compatibility with JVMs prior to this,
+resolving SocketAddresses and SocketChannels is pluggable in both client and server.
+
+### Client of a Unix Domain Socket
+
+To talk HTTP via a Unix Domain Socket at `/tmp/test.sock`, create your own implementations of `:address-finder` and `:channel-factory` in a customised client:
+
+```clojure
+(let [client (client/make-client {:address-finder (fn [uri]
+                                                    (UnixDomainSocketAddress/of "/tmp/test.sock"))
+                                  :channel-factory (fn [address]
+                                                      (SocketChannel/open StandardProtocolFamily/UNIX))})]
+   (client/get "http://foobar"))
+```
+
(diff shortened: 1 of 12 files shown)


kipz:
@ptaoussanis @borkdude - what do you reckon?

README.md
```diff
@@ -53,9 +53,38 @@ This default may be changed in a future breaking release. In the meantime, manua
   (binding [org.httpkit.client/*default-client* sni-client/default-client]
     <...>)
 ```
-
 See `org.httpkit.client/*default-client*` docstring for more details.
 
+## Using Unix Domain Sockets in client and server
+
+Java native Unix Domain Sockets are available since Java 17, so to maintain backwards compatibility with JVMs prior to this,
```
ptaoussanis, line 60: Just to double check: it looks like UDS support may have been available from Java 16 via [JEP-380](https://openjdk.org/jeps/380). Is it possible http-kit + UDS might work from Java >= 16?

README.md
```diff
@@ -53,9 +53,38 @@ This default may be changed in a future breaking release. In the meantime, manua
...
+### Client of a Unix Domain Socket
+
+To talk HTTP via a Unix Domain Socket at `/tmp/test.sock`, create your own implementations of `:address-finder` and `:channel-factory` in a customised client:
+
+```clojure
+(let [client (client/make-client {:address-finder (fn [uri]
+                                                    (UnixDomainSocketAddress/of "/tmp/test.sock"))
+                                  :channel-factory (fn [address]
+                                                      (SocketChannel/open StandardProtocolFamily/UNIX))})]
+   (client/get "http://foobar"))
```
ptaoussanis, line 72: I may be missing something obvious here, but this example looks strange.

The client in `client/get` is referring to the `HttpClient` in the `let`.
Was the intention maybe to call `org.httpkit.client/get` with `{:client <client-in-let>}` opts?

ptaoussanis (commented):
Couple initial questions before I can get this to a REPL.

README.md
ptaoussanis: Assuming Java 16 for now, can fix README later if I've missed something 👍

README.md
ptaoussanis: Checked at the REPL and made some suggested tweaks to the examples. Marking as resolved 👍

ptaoussanis:
Switched around some other priorities so that I could look at this now, PR looks good to me - have merged 👍

Thanks again for the awesome work on this James 🚀

PR #511, opened 2023-03-03
[new] [Build] [#509] Add native-image test to CI (@borkdude)

### LINKED ISSUE #509: Add native-image test to CI
@ptaoussanis Now that we have an actively running Github action, I think we should add a native image test to prevent regressions in that area. I'm willing to do a PR.

borkdude:
This commit adds:

* a test-native project in which httpkit client + server is tests with GraalVM native-image
* a deps.edn which is used by test-native but can be used by any deps.edn user
* a Github Action yml for running test-native in linux, macOS and Windows
* reflection fixes in httpkit server which were found during the tests

Fixes #509

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 *.iml
-.clj-kondo/
-.lsp/
+.cache
 .calva/
 *jar
 .classpath
@@ -15,3 +14,7 @@ pom.xml*
 /target
 /.idea
 /_site
+native-test-0.0.1-standalone
+native-test-0.0.1-standalone.build_artifacts.txt
+target
+.lein-repl-history
--- a/src/org/httpkit/server.clj
+++ b/src/org/httpkit/server.clj
@@ -7,6 +7,8 @@
            [java.nio.channels ServerSocketChannel]
            java.security.MessageDigest))
 
+(set! *warn-on-reflection* true)
+
 ;;;; Ring server
 
 (defprotocol IHttpServer
@@ -99,17 +101,22 @@
               legacy-return-value? true
               server-header "http-kit"}}]]
 
-  (let [err-logger
+  (let [^ContextLogger err-logger
         (if error-logger
           (reify ContextLogger (log [this message error] (error-logger message error)))
           (do    ContextLogger/ERROR_PRINTER))
 
-        evt-logger
+        ^ContextLogger warn-logger
+        (if warn-logger
+          (reify ContextLogger (log [this message error] (warn-logger message error)))
+          HttpServer/DEFAULT_WARN_LOGGER)
+
+        ^EventLogger evt-logger
         (if event-logger
           (reify EventLogger (log [this event] (event-logger event)))
           (do    EventLogger/NOP))
 
-        evt-names
+        ^EventNames evt-names
         (cond
           (nil?                 event-names)  EventNames/DEFAULT
           (map?                 event-names) (EventNames. event-names)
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -1,6 +1,5 @@
 *.iml
-.clj-kondo/
-.lsp/
+.cache
```
borkdude, line 2: Both clj-kondo and clojure-lsp use a `.cache` directory, that this the only part that should be ignored, the rest is configuration.

.gitignore
```diff
@@ -15,3 +14,7 @@ pom.xml*
 /target
 /.idea
 /_site
+native-test-0.0.1-standalone
```
borkdude, line 17: This is output from native-image during compilation.

src/org/httpkit/server.clj
```diff
@@ -7,6 +7,8 @@
            [java.nio.channels ServerSocketChannel]
            java.security.MessageDigest))
 
+(set! *warn-on-reflection* true)
```
borkdude, line 10: I added a lot of type hints to avoid reflection and also enabled `*warn-on-reflection* to catch those better.

test-native/deps.edn
```diff
@@ -0,0 +1,6 @@
+{:deps {httpkit/httpkit {:local/root ".."}}
```
borkdude, line 1: I chose deps.edn to depend on httpkit locally to be able to test changes quickly.
I didn't know how to do that with leiningen.

test-native/bb.edn
```diff
@@ -0,0 +1,24 @@
...
+        :task (when (seq (fs/modified-since jar ["src"]))
+                (clojure "-T:build uber"))}
+  native:build {:doc "Compile native image"
+                :depends [uber]
+                :task (when (seq (fs/modified-since native jar))
+                        (shell "native-image"
+                               "--initialize-at-build-time=clojure,org.httpkit,native"
+                               "--no-fallback"
+                               "-jar" jar))}
+  native:test {:doc "Run native test"
```
borkdude, line 22: This kicks off the whole chain of build steps necessary to build the native image and then runs it.

borkdude:
@ptaoussanis Added a few clarifications as comments. Let me know if you have more questions.

borkdude:
@kipz FWIW, locally the tests work fine for me with `lein test :gha`

src/org/httpkit/server.clj
ptaoussanis: Is the location of this intentional?
Won't this override the user's intended `*warn-on-reflection*` on ns eval?

ptaoussanis:
PR looks good to me, thanks @borkdude! 🙏

> @kipz Any idea why the "build-and-test" workflow is stuck? I enabled it for PRs but maybe I shouldn't have done that?

Is it possible the failure may have been transient (or is at least inconsistent)?

Requested a re-run now and it seemed to [pass okay](https://github.com/http-kit/http-kit/actions/runs/4327933034/jobs/7561663834).

But will hold off merging until you both indicate that you're happy with the PR as is.

src/org/httpkit/server.clj
borkdude: It will always enable reflection warnings in the server namespace, which is what we want to always prevent, right?
It doesn't have an effect on other namespaces.

borkdude:
@ptaoussanis I'm ok with this PR as is. Slightly worried about the flakiness of the `lein test :gha` but that isn't related to this PR and we'll just keep an eye on that.

src/org/httpkit/server.clj
ptaoussanis: Sorry - I might be missing something obvious here, but `clojure.core/*warn-on-reflection*` is a global var - how would setting it not affect other namespaces?

borkdude:
@ptaoussanis Added one more commit. This tests for reflection issues before compiling a native, which is usually the culprit with native image.

src/org/httpkit/server.clj
ptaoussanis: Interesting, I see I'm mistaken! The [`set!` doc](https://clojure.org/reference/vars#set) indeed clarifies that it's only a thread-local effect.

I've somehow never encountered this before, that's actually very handy to know!

Thanks for pointing this out :-)

borkdude: You're welcome :)

kipz:
@borkdude yeah, there must be a flaky test in there. Agree it has nothing to do with this PR.

kipz (approved):
LGTM

borkdude:
@ptaoussanis  I'll wait for you to merge unless you tell me it's ok to merge myself, as I don't want to disrupt your normal style of commit messages, etc. :)

ptaoussanis:
@borkdude @kipz Thanks both, I've merged 👍

> I don't want to disrupt your normal style of commit messages, etc. :)

Yours was perfect, much appreciated :-)

Issue #393, opened 2019-02-19
SNI support to be enabled by default?


huima:
If I would make a pull request which would enable the SNI support in http-kit client by default, is there any reasons why it would not be accepted? Are there any security or other reasons to keep the current config where SNI needs to be enabled by own ssl configurator.

We are using http-kit heavily and now need to configure SNI support to be enabled in every project.

kumarshantanu:
AFAIK support for SNI requires Java 8, and making that a requirement would leave many Java 6/7 users out. Linking old issues on this subject: #187 #311 #335

huima:
Yup. It could have a check via java interop to check that the java.vm.version system property is 1.8 ->

jwr:
I've just spent the better part of a day debugging this problem and ended up here. I think the current situation is untenable: sites proxied via CloudFlare require SNI, as do many other places on the Internet.

I think support for older JVM versions is secondary. These days code running on Java 6 and 7 will not be able to connect to many modern APIs or sites anyway, because of 1) SNI, and 2) limited crypto policy that is enforced by default, so without manual installation of JCE unlimited policy files many sites will fail because of cipher negotiation.

I implemented the workaround suggested by @kumarshantanu (thanks!), but I think http-kit client is deficient right now and the lack of SNI is a major pitfall which deserves at least a mention in the README.

ptaoussanis:
Hey everyone, sorry for the inactivity on this- haven't had an opportunity to spend any time on this.

Could someone kindly summarise the current situation, options, and trade-offs?
Is there a consensus on a particular action to take?

jwr:
Well, I've outlined my rationale above — not sure what I could add to that. To rephrase and summarize:

1. SNI is required for many modern sites & APIs (it seems anything hosted behind Cloudflare requires it now).

2. Java 6/7 does not support SNI, so these apps will not work with modern sites & APIs anyway.

3. Many modern sites also require modern crypto like AES-256, which in Java 6 and 7 requires installing additional libraries to unblock the JCE unlimited crypto policy. This means that old Java 6/7 apps will not work anyway without changes to the JVM installation.

My conclusion is that while I love backwards compatibility, in this case I think it makes little sense, at least for the http-client part. Perhaps a last release that supports Java 6/7 could be made, and then we should start requiring Java 8 and support SNI.

huima:
Totally agree with @jwr 

[...]

PR #513, opened 2023-03-07
[new] [Client] [#393] Use SNI client by default for Java >= 8

ptaoussanis:
Intended to solve https://github.com/http-kit/http-kit/issues/393.

Based on [work](https://github.com/aiba/http-kit/commit/b31625fac0402732cb0b41b5e6cd01ce5d9fa385) by @aiba, this PR will use the http-kit SNI client by default on Java >= 8.

This change is intended to be non-breaking, and won't bump the minimum Java requirement.
Once this PR is in, we can decide later about if/when we'd like to just remove the old client altogether.

### Current PR Status: BLOCKED

It seems that defaulting to SNI client currently breaks a a number of existing [client tests](https://github.com/http-kit/http-kit/blob/master/test/org/httpkit/client_test.clj).

**Merge blocked** on fixing these tests.
Assistance very welcome, otherwise I'll try take a closer look myself in a few weeks.


src/org/httpkit/client.clj
```diff
@@ -185,6 +163,22 @@ an SNI-capable one, e.g.:
 
    bind-address))
 
+;; Normally only need one client per application - params can be configured per req
+(defonce default-client
+  (delay
+    (if (utils/java-version>= 8)
+      (do ; Note "Gilardi scenario"
+        (require 'org.httpkit.sni-client)
```
borkdude, line 171: @ptaoussanis Doing a `require` and `ns-resolve` in a delay will mess with graalvm. 

It _might_ _work_, but it will blow up the image size unnecessarily. 

If you would allow me, I'd have a better alternative for this, by moving some of the require + resolve to the top level.

borkdude: And on the other hand, why not just drop support for Java 7. Clojure has been doing this since ages now.

ptaoussanis: @borkdude 

> Doing a require and ns-resolve in a delay will mess with graalvm.

Thanks for considering this. Is this something the recently added [tests](https://github.com/http-kit/http-kit/pull/511) should catch? They seem to pass okay with the change? (I can't speak for image size)

> It might work, but it will blow up the image size unnecessarily

Is it possible to briefly explain why (or point me to something I can read?) - would be curious to understand better why this'd trigger a significant image size increase.

> And on the other hand, why not just drop support for Java 7. Clojure has been doing this since ages now.

I wasn't aware of any downside to continued support, and I'm aware of at least a couple orgs that still use http-kit on Java 7 in production.

I'm not opposed to dropping support if support has non-trivial costs (e.g. issues for GraalVM users).

borkdude: The native image tests already require the namespace beforehand, this is why they currently work. If that wouldn't have been done, the tests would likely fail.

As for "image bloat": doing requires and resolves at runtime (other than top level) will increase babasha from 75mb to 120mb. The reason for this I'm not exactly sure on, but it has to do with the reachability analysis that GraalVM does. If it can't decide that something is unused, it will leave things in.

ptaoussanis: @borkdude Assuming the issue for GraalVM is the lazy require+resolve, is [this](https://github.com/http-kit/http-kit/pull/513/commits/be608e85440a9edfab11360b548a7b07f7796dd6) better?

src/org/httpkit/client.clj
```diff
@@ -163,14 +163,17 @@
 
    bind-address))
 
+(def ^:private ssl-configurer
```
borkdude, line 166: This looks great, thanks.

src/org/httpkit/client.clj
borkdude: Yep, that solves it. Perhaps doing `(some-> (ns-resolve ..) deref)` would be a bit better even, but not sure if that matters.

ptaoussanis: PR updated 👍 Thanks again!

ptaoussanis:
Identified the cause of the test breakage, and looks like it'd be reasonable to temporarily disable the affected tests for forthcoming v2.7 alpha.

Have opened a separate issue (https://github.com/http-kit/http-kit/issues/519) for reenabling the tests, so the present issue can be closed.

Issue #137, opened 2014-03-29
Server hangs with chunked transfer and a length set to -1


ramanb:
POST / HTTP/1.1
Host: 127.0.0.1
Transfer-Encoding: chunked

-100

Server hangs, logs spit

Fri Mar 28 14:36:44 EST 2014 [server-loop] ERROR - http server loop error, should not happen
java.lang.NegativeArraySizeException
        at org.httpkit.server.HttpDecoder.decode(HttpDecoder.java:95)
        at org.httpkit.server.HttpServer.decodeHttp(HttpServer.java:100)
        at org.httpkit.server.HttpServer.doRead(HttpServer.java:164)
        at org.httpkit.server.HttpServer.run(HttpServer.java:277)
        at java.lang.Thread.run(Thread.java:744)
Parameter Pollution

ptaoussanis:
Could someone confirm if this is still an issue on master?

kipz:
@ptaoussanis I have a failing test for this one. I'll see what I can do to fix.

kipz:
On the face of it, this problem looks a bit wider than I thought. There are many things that we can do to the HTTP request to break main loop and stop the channel from being closed. AFAICT, there's no maximum timeout either, so just a normal (non-chunked) request with a Content-Length > 1 and no body leads to hanging. At the moment, I'm trying to unpick connection re-use from plain hanging channels....

PR #514, opened 2023-03-07
Close idle channels after 1000 millis

kipz:
Not ready for review. Just an experiment at this point.

Relates to https://github.com/http-kit/http-kit/issues/137

Turns out there are many different invalid http requests we can make that cause the server to never complete the request. E.g.

* Send a negative Content-Length and no chunks in a chunked request
* Send an empty body but include a positive Content-Length

This is an attempt to close channels that end up in this state after 1s of inactivity.

TBH - I don't fully know the implications of this change yet, especially on connection reuse.

### DIFF
--- a/src/java/org/httpkit/server/HttpServer.java
+++ b/src/java/org/httpkit/server/HttpServer.java
@@ -292,6 +292,8 @@ public class HttpServer implements Runnable {
 
     private void doRead(final SelectionKey key) {
         SocketChannel ch = (SocketChannel) key.channel();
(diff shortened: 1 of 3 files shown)


harold:
Would this affect uses of `as-channel` that sat idle for more than 1s? This is a use case for us for infrequent 'server sent events'.

ptaoussanis:
Closing as part of issue triage, please feel free to reopen if anyone is interested in contributing further efforts on this.

PR #524, opened 2023-06-08
Reliably close InputStream

rublag:
The `readAll` function reads all data from a stream. If the size of the data is too large, JVM will throw `OutOfMemoryError`. In this case `is.close()` will not be called which could lead to resource leak.


ptaoussanis:
@rublag Thanks for this! Will include with the next release 👍

ptaoussanis:
Merged manually, thanks again @rublag! 🙏

PR #531, opened 2023-08-28
Update http.async.client to 1.2.0

### LINKED ISSUE #530: Outdated http.async.client fails to compile
Hey friends,

http.async.client, a dependency in the `:dev` profile, fails to compile when used with a version of Clojure later than 1.8. [Here's the PR where it's fixed.](https://github.com/cch1/http.async.client/pull/76) This doesn't matter for downstream users of http-kit as it's a dev-only dependency, but it makes it harder to work on http-kit if we can't use recent versions of Clojure (to verify things like compilation). Updating to version 1.2.0 should be the earliest version that includes the fix.

Thanks so much.

NoahTheDuke:
Closes #530

CI doesn't check against all of the versions of Clojure, so there could be future regressions. I don't know what your desired CI workflow is like, so I'll leave that up to you.


ptaoussanis:
Merging manually, thanks Noah! 🙏

NoahTheDuke:
I'm glad to help!

Issue #533, opened 2023-09-27
Capture number of processors at runtime in a graalvm binary


borkdude:
The number of processors are captured in a top level value: 

https://github.com/http-kit/http-kit/blob/ff5802d8a7c59ff84894a0d4ebd5eff775bedb2e/src/org/httpkit/client.clj#L89

https://github.com/http-kit/http-kit/blob/ff5802d8a7c59ff84894a0d4ebd5eff775bedb2e/src/org/httpkit/server.clj#L47

In a GraalVM binary this value is captured at compile time. It'd be better to delay this expression to runtime so it truly captures the number of processors on the user's machine and not that of the build environment.

Happy to follow up with a PR.

borkdude:
PR here: https://github.com/http-kit/http-kit/pull/534

ptaoussanis:
I believe this can be closed now 👍

Ended up taking a slightly different approach since I was anyway planning to introduce some worker utils which also bypass the problem.

Thanks again for the report and work on this! Cheers :-)

PR #534, opened 2023-09-27
[fix] [#533] Capture number of processors at runtime in graalvm binary

borkdude:


borkdude:
I'm going to look into why the graal tests are failing

ptaoussanis:
> I'm going to look into why the graal tests are failing

I suspect the var used for tests isn't anticipating a delay.
I'm happy to handle this myself tomorrow.

Thanks again!

borkdude:
@ptaoussanis Already fixed. The Graal issue had to do with reflection, adding `(int ..)` solved it.

Issue #535, opened 2023-09-28
http-kit client accepts some bad SSL certificates by default


jeffdik:
The http-kit client from versions 2.4.0-alpha1 to 2.8.0-beta1 allow some bad SSL certificates by default. For example, no exception is thrown doing a HTTP GET to `https://expired.badssl.com/`, `https://self-signed.badssl.com/`, or `https://untrusted-root.badssl.com/`.

Here's an example of a HTTP GET to https://expired.badssl.com/ with http-kit 2.4.0-alpha1:
```
> clojure -Sdeps '{:deps {http-kit/http-kit {:mvn/version "2.4.0-alpha1"}}}' -M -e '(require \'org.httpkit.client) @(org.httpkit.client/get "https://expired.badssl.com/")'
{:opts {:method :get, :url "https://expired.badssl.com/"}, :body "<!DOCTYPE html>\n<html>\n<head>\n  <meta charset=\"utf-8\">\n  <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\">\n  <link rel=\"shortcut icon\" href=\"/icons/favicon-red.ico\"/>\n  <link rel=\"apple-touch-icon\" href=\"/icons/icon-red.png\"/>\n  <title>expired.badssl.com</title>\n  <link rel=\"stylesheet\" href=\"/style.css\">\n  <style>body { background: red; }</style>\n</head>\n<body>\n<div id=\"content\">\n  <h1 style=\"font-size: 12vw;\">\n    expired.<br>badssl.com\n  </h1>\n</div>\n\n</body>\n</html>\n", :headers {:date "Tue, 26 Sep 2023 16:16:40 GMT", :server "nginx/1.10.3 (Ubuntu)", :etag "W/\"64bafff7-1ee\"", :last-modified "Fri, 21 Jul 2023 22:00:23 GMT", :cache-control "no-store", :content-type "text/html", :content-encoding "gzip", :connection "keep-alive", :transfer-encoding "chunked"}, :status 200}
```

http-kit 2.3.0 throws a SSLHandshakeException:
```
> clojure -Sdeps '{:deps {http-kit/http-kit {:mvn/version "2.3.0"}}}' -M -e '(require \'org.httpkit.client) @(org.httpkit.client/get "https://expired.badssl.com/")'
{:opts {:method :get, :url "https://expired.badssl.com/"}, :error #error {
 :cause "unable to find valid certification path to requested target"
 :via
[...]

PR #536, opened 2023-09-28
[fix] [client] fix handling of some bad ssl certificates

jeffdik:
Broke in e28e4c6e77835d2768a117f12a4ab3e3a42283b2 (between v2.3.0 and v2.4.0-alpha1)

### DIFF
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -89,9 +89,7 @@ public class HttpClient implements Runnable {
     public static SSLContext getDefaultContext() {
         if(defaultContext==null) {
             try {
-                SSLContext sslContext = SSLContext.getInstance("TLS");
-                sslContext.init(null, TrustManagerFactory.getTrustManagers() ,null);
-                defaultContext = sslContext;
+                defaultContext = SSLContext.getDefault();
             } catch (Exception e) {
                 throw new Error("Failed to initialize SSLContext", e);
             }
--- a/test/org/httpkit/client_test.clj
+++ b/test/org/httpkit/client_test.clj
@@ -360,6 +360,20 @@
   (is (contains? @(hkc/get "https://microsoft.com")   :status))
   (is (contains? @(hkc/get "https://letsencrypt.org") :status)))
 
+(deftest test-bad-ssl-certs
+  (doseq [url ["https://expired.badssl.com/"
+               "https://wrong.host.badssl.com/"
+               "https://self-signed.badssl.com/"
+               "https://untrusted-root.badssl.com/"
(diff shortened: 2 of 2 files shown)


jeffdik:
Potential fix for https://github.com/http-kit/http-kit/issues/535

ptaoussanis:
@jeffdik Hi Jeff, thanks so much for the report and fix - this looks great! 🙏

What's the current status on this PR? 
You mentioned that you believe this is a "potential" fix for #535. Is there still any additional testing/input that you're looking for, or is this maybe ready for inclusion in the next beta?

(I'm not familiar with the implementation of the SSL system myself, so unfortunately can't give any specific guidance).

jeffdik:
Hi @ptaoussanis, thanks so much for all your work maintaining this project!

I think this PR is ready for inclusion in the next beta. All the tests passed for me.

I called it a "potential" fix because I'm not really familiar with SSL/TLS or Java's way to to use SSL/TLS or http-kit's implementation and was hoping that someone with more knowledge could look at it in case I'm missing something.

I think the things that the fix has going for it is that it's what http-kit used to do (2.3.0 and before) and that it's what [org.httpkit.client-test/test-default-sni-client](https://github.com/http-kit/http-kit/blob/ff5802d8a7c59ff84894a0d4ebd5eff775bedb2e/test/org/httpkit/client_test.clj#L372) does via [make-ssl-engine](https://github.com/http-kit/http-kit/blob/ff5802d8a7c59ff84894a0d4ebd5eff775bedb2e/src/org/httpkit/client.clj#L115). I took a brief look at the [SSLContext Class](https://docs.oracle.com/en/java/javase/21/security/java-secure-socket-extension-jsse-reference-guide.html#GUID-C281CAF3-275F-4DE4-8B47-4A84363CF39F) section of [Java Secure Socket Extension (JSSE) Reference Guide](https://docs.oracle.com/en/java/javase/21/security/java-secure-socket-extension-jsse-reference-guide.html) and it looks like calling the SSLContext.getDefault method is reasonable, but I'm very much a newbie in this area.

I also noticed [sslContext.init(null, TrustManagerFactory.getTrustManagers() ,null);](https://github.com/http-kit/http-kit/blob/02a373939e4fdcc545329d6969c4a792ab7fdb32/src/java/org/httpkit/client/HttpClient.java#L93) calls `org.httpkit.client.TrustManagerFactory` which creates a [DUMMY_TRUST_MANAGER](https://github.com/http-kit/http-kit/blob/02a373939e4fdcc545329d6969c4a792ab7fdb32/src/java/org/httpkit/client/ClientSslEngineFactory.java#L42-L60) which stubs out `checkClientTrusted` and `checkServerTrusted` methods which only contain the comment "Always Trust", which doesn't seem right.

ptaoussanis:
You're very welcome Jeff!

Likewise thanks for the very clear info, and solid PR - it looks good to me.
Have merged your fix manually, it'll be included for wider testing in the next beta.

Really nice work, cheers :-)

PR #538, opened 2023-09-28
[new] [server] [#394] Add support for Ring async handler arities

### LINKED ISSUE #394: Support ring async?
It would be awesome if http-kit server could support the ring async handler mechanism.

```clojure
(defn handler
  ([request]
   {:status 200, :headers {}, :body "Hello World"})
  ([request respond raise]
   (respond {:status 200, :headers {}, :body "Hello World"}))
```

Right now it appears to always use the synchronous mechanism:

```java
Map resp = (Map)this.handler.invoke(ClojureRing.buildRequestMap(this.req));
```

I haven't tested, but I suspect this could be tweaked and if `this.handler` is of type `MultiFn`, it could invoke the 3 arity handler instead and the `HttpHandler` thread would have to exit so it can be truly async and not block a request thread.

weavejester:
This patch adds support for the three-arity async handlers introduced in Ring 1.6, and closes #394. The `:ring-async?` option, off by default, toggles support for async Ring handlers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,12 +32,11 @@ This will build the java source and test files, it will then execute 'lein test'
 
 To run the java junit tests, at the root of the project type:
 
-```rake junit```
+```rake test_java```
 
 To run a set of benchmarks, at the root of the project type:
 
-```rake benchmark```
-
+```rake bench```
 
 ## Filing Pull Requests
 
--- a/Rakefile
+++ b/Rakefile
@@ -1,40 +1,16 @@
 task :default => :test
 
-desc "Run unit test"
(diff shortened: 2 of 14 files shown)


weavejester:
It looks like `with-timeout` is failing to eval its `timeout-form` under some circumstances, resulting in the `test-timer` test hanging. There may be some race condition with the `TimerService` class.

weavejester:
I've identified the race condition and submitted PR #540. Once that's merged, the CI for this PR should succeed consistently.

ptaoussanis:
@weavejester Merged manually, this'll be awesome to have in - thank you so much! Cheers :-)

Issue #539, opened 2023-09-29
Race condition in TimerService.scheduleTask


weavejester:
Under some circumstances, `TimerService.scheduleTask` can hang until another task is scheduled. This is caused when the timer service thread ends inbetween the `started` variable being checked, and the new task being queued:

```java
    public CancelableFutureTask scheduleTask(int timeout, IFn task) {
        // if the timer-service thread is started here
        if (started.compareAndSet(false, true)) {
            Thread t = new Thread(this, "timer-service");
            t.start();
        }

        // but stopped here, then the task will be queued but never executed until another task is scheduled
        CancelableFutureTask t = new CancelableFutureTask(timeout, task, queue);
        synchronized (queue) {
            queue.offer(t);
            queue.notify();
        }
        return t;
    }
```

I'll create a PR to fix this.

ptaoussanis:
@weavejester Merged manually, thanks so much for the clear info + PR James, cheers! :-)

borkdude:
Nice find + fix @weavejester !

PR #540, opened 2023-09-29
[fix] [server] [#539] Prevent race condition in TimerService

weavejester:
This PR fixes the race condition identified in #539 by moving the timer-service thread into the synchronized block. It's placed after the task is placed on the queue to guarantee that it gets a task on its first loop.

I also discovered that the `queue.wait()` for empty queues also needed to be moved, otherwise the `queue.notify()` called from `scheduleTask` could trigger an immediate break of the loop. By moving queue.wait() to the else-branch, it ensures that notifying the queue object will always hit the top of the while loop in the timer-service thread.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,12 +32,11 @@ This will build the java source and test files, it will then execute 'lein test'
 
 To run the java junit tests, at the root of the project type:
 
-```rake junit```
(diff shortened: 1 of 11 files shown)


ptaoussanis:
Merged manually, forgot to close this earlier 👍 🙏 🙏

Issue #543, opened 2023-10-07
Problem in native-image and HttpUtils when using virtual threads


borkdude:
When I run the below example with babashka:

``` clojure
(require '[org.httpkit.server :as server])

(defn app [_]
  {:status 200
   :body "Hello HTTP!"
   :headers {"Content-Type" "text/plain"}})

(server/run-server app {:port 8080
                        :worker-pool
                        (java.util.concurrent.Executors/newVirtualThreadPerTaskExecutor)
                        #_(java.util.concurrent.Executors/newFixedThreadPool 8)})

(println "Server started at http://localhost:8080")

@(promise) ;; wait for SIGINT
```

and I bombard the server with many requests like in this blogpost:

https://lemire.me/blog/2023/10/07/web-server-hello-world-benchmark-go-vs-node-js-vs-nim-vs-bun/
(also see https://twitter.com/lemire/status/1710538462460256437)

I'm getting to see:

```
java.util.NoSuchElementException
	at java.base@21/java.util.ArrayDeque.removeFirst(ArrayDeque.java:361)
	at java.base@21/java.util.ArrayDeque.pop(ArrayDeque.java:592)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.PrivilegedStack.pop(AccessControllerUtil.java:88)
	at java.base@21/java.security.AccessController.executePrivileged(AccessController.java:131)
	at java.base@21/java.security.AccessController.doPrivileged(AccessController.java:319)
	at java.base@21/sun.util.resources.LocaleData.getBundle(LocaleData.java:182)
	at java.base@21/sun.util.resources.LocaleData.getDateFormatData(LocaleData.java:145)
	at java.base@21/java.text.DateFormatSymbols.initializeData(DateFormatSymbols.java:747)
	at java.base@21/java.text.DateFormatSymbols.<init>(DateFormatSymbols.java:151)
	at java.base@21/sun.util.locale.provider.DateFormatSymbolsProviderImpl.getInstance(DateFormatSymbolsProviderImpl.java:85)
	at java.base@21/java.text.DateFormatSymbols.getProviderInstance(DateFormatSymbols.java:371)
	at java.base@21/java.text.DateFormatSymbols.getInstanceRef(DateFormatSymbols.java:361)
	at java.base@21/java.text.SimpleDateFormat.<init>(SimpleDateFormat.java:626)
	at org.httpkit.DateFormatter.initialValue(HttpUtils.java:26)
[...]

PR #544, opened 2023-10-07
[fix] [server] [#543] Migrate away from SimpleDateFormat to java.time

borkdude:
Fixes issue #543, problem in native-image with ThreadLocal + SimpleDateFormat + virtual threads.


ptaoussanis:
Merged manually, thanks again!

PR #546, opened 2023-10-11
[WIP] Add support for Ring websocket API

### LINKED ISSUE #541: Ring websocket API support
I've been working on a [Ring websocket API](https://github.com/ring-clojure/ring/blob/master/SPEC-alpha.md#3-websockets) that's currently in experimental alpha. Would you be interested in a PR to support this in http-kit, to be merged when the Ring websocket API is stable? It would obviously be supported in conjunction with http-kit's native implementation.

The reason I'm asking this while the Ring API is still experimental is that I'd like to try implementing the API for different third-party adapters, so I can get feedback and forewarning of any design issues. I've submitted [a similar issue](https://github.com/sunng87/ring-jetty9-adapter/issues/115) to the ring-jetty9-adapter project.

weavejester:
This PR **should not** yet be merged, as the Ring websocket API is not stable. However, I wanted to implement the alpha websocket API for different adapters in order to head off any difficulties or flaws in the API design.

Implementing this for http-kit proved to be more difficult than I expected, as http-kit abstracts things like the close handler, and the Ring API is a little more low level. As an overview:

- I added a `pongHandler` to handle pong frames
- I added a `closeRingHandler` to handle close frames and pass the raw code and reason data
- I added functionality to send ping and pong frames explicitly via `send`
- I used the hato library for testing the Ring websockets in the tests, as the async.http.client one didn't support custom codes and reason strings in close frames.

I use some private middleware to detect Ring websocket responses and convert them into a http-kit channel.

This PR will eventually close #541.


weavejester:
I think the failure in the main test is unrelated to my changes. It looks like the `test-server-status` test is expecting the server to be `:stopped` when it's currrently `:stopping`, so maybe a slowdown on the CI server caused the test to fail.

ptaoussanis:
Hi James, thanks for this! Will try take a closer look tomorrow, but in the meantime wanted to respond to this-

> I think the failure in the main test is unrelated to my changes. It looks like the test-server-status test is expecting the server to be :stopped when it's currrently :stopping, so maybe a slowdown on the CI server caused the test to fail.

Correct, the failure's very possibly unrelated. Looks like the timing for that test may not leave much margin for error if CI is slow, I'll bump the allowed time.

ptaoussanis:
> This PR should not yet be merged, as the Ring websocket API is not stable. However, I wanted to implement the alpha websocket API for different adapters in order to head off any difficulties or flaws in the API design.

👍 Thanks for the clear status, and all your work on this James!

Took a quick skim at the current PR, looks broadly reasonable from my end.

Only concern that stuck out - it looks like your `wrap-ring-handler` fn needs access to `ring.websocket/Socket`, which'll mean a dependency on `ring/ring-core`?

Is access the `Socket` protocol the only reason for the dependency? If so, would there be any way to avoid that? 

One of http-kit's major value propositions is it's lack of dependencies.

weavejester:
This is exactly the sort of feedback I was hoping to get, as it's something I wouldn't have considered otherwise. So first of all, thank very much for taking the time to review this.

What if I split the relevant protocols out into their own library, so I'd have a `org.ring-clojure/ring-websocket-protocols` library, that would contain *only* the following namespace and nothing else:

```clojure
(ns ring.websocket.protocols)

(defprotocol Listener
  (on-open [listener socket])
  (on-message [listener socket message])
  (on-pong [listener socket data])
  (on-error [listener socket throwable])
  (on-close [listener socket code reason]))

(defprotocol PingListener
  (on-ping [listener socket data]))

(defprotocol Socket
  (-open? [socket])
  (-send [socket message])
  (-ping [socket data])
  (-pong [socket data])
  (-close [socket code reason]))

(defprotocol AsyncSocket
  (-send-async [socket message succeed fail]))
```

While this would still add a dependency, it would be as minimal as possible with no code beside the relevant protocols. Would this be acceptable?

borkdude:
Seems like an idea. I wonder if there's ways to do it without adding dependencies at all. "Just use maps"? Or maybe that would be too brittle?

borkdude:
I guess another idea would be to add:

1 a protocol to http-kit
2 a separate dependency which bridges http-kit and ring-core-websockets

Maybe that's too much overkill though

borkdude:
Wait a minute, does this PR mean that httpkit's websocket server won't work at all anymore without ring-core as a dependency, whereas it previously did work? Several people are relying on this feature in babashka right now.

I can check with this PR how it influences babashka though. Maybe it's not too bad.

borkdude:
A note: ring-core pulls in more transitive dependencies as well:

```
ring/ring-core 1.11.0-alpha4
  . ring/ring-codec 1.2.0
  . commons-io/commons-io 2.13.0
  . org.apache.commons/commons-fileupload2-core 2.0.0-M1
    . commons-io/commons-io 2.13.0
  . crypto-random/crypto-random 1.2.1
    . commons-codec/commons-codec 1.15
  . crypto-equality/crypto-equality 1.0.1
```

Maybe another good argument to split things up.

My question would be: would the existing websocket functionality (as of before this PR) keep on working, when you do not include the full `ring-core` dependency? (If not, that would be breaking, e.g. current babashka users)

weavejester:
> Seems like an idea. I wonder if there's ways to do it without adding dependencies at all. "Just use maps"? Or maybe that would be too brittle?

We could do something like this, where both the listener and socket are maps of functions:

```clojure
{:ring.websocket/listener
 {:on-open (fn [{:keys [send ping open?]}]
             (when (open?)
               (ping) 
               (send "Hello World"))}}
```

The problem is that this would be less performant than using protocols, the functions wouldn't be enforced by the compiler, and it feels less idiomatic overall. I think my preference would still lean toward having a minimal protocol library.

> My question would be: would the existing websocket functionality (as of before this PR) keep on working, when you do not include the full `ring-core` dependency? (If not, that would be breaking, e.g. current babashka users)

Currently no, because `org.httpkit.server` requires `ring.websocket`. However, we only need `ring-core` for the Ring websocket protocols, so if we pull them out, no `ring-core` dependency would be needed.

borkdude:
Just to be clear, websocket functionality will keep on working as it is right now, without adding a dependency on ring-core, but just a small dependency on a protocol-only dependency? Sounds good enough to me!

borkdude:
With "now" I mean as things are on master, not as they are in this PR.

weavejester:
> Just to be clear, websocket functionality will keep on working as it is right now (on master), without adding a dependency on ring-core, but just a small dependency on a protocol-only dependency? Sounds good enough to me!

Yes, exactly.

weavejester:
For the socket, we could use a metadata protocol:

```clojure
(defn- channel->ring-socket [^AsyncChannel ch]
  (with-meta
   {:channel ch}
   {'ring.websocket.protocols/open? (fn [_] (not (.isClosed ch)))
    ...}))
```

I haven't tried it before, but I believe this should work. And for the listener, we could use a var lookup:

```clojure
(defn- channel->listener [^AsyncChannel ch listener]
  (require 'ring.websocket.protocols)
  (let [on-message (var-get (find-var 'ring.websocket.protocols/on-message))]
    (.setReceiveHandler ch #(on-message listener ch (->ring-message %)))
    ...))
```

There would probably be some performance loss, but I'm unsure if that would matter much for something like websockets that are mostly going to be limited by I/O.

However, regardless of whether we use that, I think it's still an excellent idea to separate out the websocket protocols into their own library.

borkdude:
Please avoid "inline" `require` in functions, it does not work well with GraalVM native-image (run-time requires don't work + run-time requires yield insanely big images because GraalVM cannot analysis that stuff isn't used). Probably better to use `utils/compile-if` to do this work at compile time so `require` isn't called anymore at runtime.

weavejester:
So something like:

```clojure
(defn- wrap-ring-websocket [handler]
  (utils/compile-if (require 'ring.websocket.protocols)
    ;; ring websocket code
    handler))
```

Would that work in GraalVM? I guess we could also use `reify` instead of metadata protocols in that case for the socket.

borkdude:
Yeah, that would work, I think.

`wrap-ring-websocket` returns nil when `ring-core` in this case. Would the existing (as now on master) websocket functionality be affected by that?

weavejester:
> `wrap-ring-websocket` returns `nil` when `ring-core` is not on the classpath in this case. Would the existing (as now on master) websocket functionality be affected by that?

Ah, sorry, my example probably wasn't clear. The `;; ring websocket code` comment was a placeholder for all the ring websocket code. If there is no ring websocket protocols included, it defaults to returning `handler` (i.e. doing nothing).

borkdude:
Gotcha!

weavejester:
I've updated the code so that Ring WebSockets are an optional dependency, and that dependency is a small library consisting of only protocols. I think the tests need a little more work to get them passing more reliably, and the Ring dependencies will need updating once we hit an RC, but overall I think it's done for now.

borkdude:
I'd be interested in one test suite which checks if the existing websocket stuff still runs without the ring dependency as well.

weavejester:
That's a good point. Maybe we just add a Leiningen profile that omits the dependency, and add that to the GitHub Actions tests.

ptaoussanis:
@weavejester Hi James, happy new year! Just confirming that this is ready to go from your side? I'm happy to update the deps and add the remaining deps-free tests when merging.

weavejester:
Yes it should be ready aside from the updated deps and the deps-free tests that you mentioned.

I also separated out the websocket protocols into a minimal `org.ring-clojure/ring-websocket-protocols` dependency, so if people want to use websockets with http-kit, but wish to avoid ring-core, that's an option for them.

PR #553, opened 2023-11-01
Fix order of  err-logger and warn-logger

### LINKED ISSUE #552: logger-warn and error-warn are passed in wrong order to HttpServer constructor
In https://github.com/http-kit/http-kit/blob/87381406e56510eaf0b27b3dfdeafcfc18b3b9d7/src/org/httpkit/server.clj#L175-L180 we pass `warn-logger` and `err-logger` to the `HttpServer` constructor in the wrong order.

The 11-arity `HttpServer` constructor expects the `err-logger` before `warn-logger`:
https://github.com/http-kit/http-kit/blob/87381406e56510eaf0b27b3dfdeafcfc18b3b9d7/src/java/org/httpkit/server/HttpServer.java#L115-L120

frwdrik:
The HttpServer constructor expects a different order of these.
Fixes #552.

### DIFF
--- a/src/org/httpkit/server.clj
+++ b/src/org/httpkit/server.clj
@@ -174,8 +174,8 @@
 
         s (HttpServer. address-finder channel-factory h
                        ^long max-body ^long max-line ^long max-ws proxy-enum ^String server-header
-                       warn-logger
                        err-logger
+                       warn-logger
                        evt-logger
                        evt-names)]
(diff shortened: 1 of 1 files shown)


ptaoussanis:
@frwdrik Thanks a lot for the report + fix Fredrik! Merged manually 👍

PR #554, opened 2023-11-02
Support for non-ASCII characters on multipart filename

davartens:
- When executing a POST request to a mail broker, file attachments which include non-ascii characters on its filename get its non-ascii characters replaced by '?'. 
- We use lots of attachments with spanish characters on its filenames. Support for non-ascii characters like 'ñ' or 'á' would be ideal.
- The suggestion is to use UTF-8 encoding for filename instead of the default ISO-8859-1.


ptaoussanis:
@davartens Great, thanks for this! Merged manually 👍

PR #569, opened 2024-05-21
[Client][Performance] Implement PersistentConn equality dynamically

### LINKED ISSUE #568: [client][performance] Regression after #446
We hit an edge case at work where 
```java
keepalives.remove(job.addr.toString() + job.host);
```
hogged the client's event loop and caused a server to fall over with relatively light load.
Looks like this change is related to #446 

Possible fix:
Define a class which will hold addr and job and defines equals when they're both equal

wdyt?

bsless:
Closes #568

### DIFF
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -450,7 +450,7 @@ public class HttpClient implements Runnable {
         Request job = pending.peek();
         if (job != null) {
             if (job.cfg.keepAlive > 0) {
-                PersistentConn con = keepalives.remove(job.addr.toString() + job.host);
+                PersistentConn con = keepalives.remove(job);
                 if (con != null) { // keep alive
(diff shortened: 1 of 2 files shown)


bsless:
Hi @ptaoussanis, any update on this? It's a pretty bad performance bug and it makes using the client at scale with multiple hosts completely unfeasible

ptaoussanis:
@bsless Hi Ben,

I've just pushed v2.9.0-alpha1 to [Clojars](https://clojars.org/http-kit/versions/2.9.0-alpha1) which includes your fix.

If you can confirm that that's stable and successfully addresses the problem, then I'll aim to cut 2.8.1 with the same fix.

Thanks for your work on this 👍

bsless:
I'll look for a scenario where I can test it

ptaoussanis:
Going to assume that this is resolved, will include in the forthcoming unstable release 👍

Issue #570, opened 2024-05-31
Add updated WebSocket examples to Wiki


ptaoussanis:
The [Wiki](https://github.com/http-kit/http-kit/wiki) could really use simple, up-to-date documentation/examples of how to effectively use WebSocket with http-kit client + server.

With the latest versions of http-kit, there are two ways to do this:

1. Using http-kit's own "AsyncChannel" API
2. Using the standard Ring WebSocket API

It'd be nice to have both explained, and contrasted

# Approach 1

A typical flow would be something like this-

- Your Ring handler uses [as-channel](https://cljdoc.org/d/http-kit/http-kit/CURRENT/api/org.httpkit.server#as-channel) to setup handlers for open/close/receive events. Those handlers will be called on activity, with an arg for the underlying http-kit AsyncChannel.

- You use an atom or something similar to keep track of active channels for server->client messaging.

In this setup, http-kit’s handler threads are used only briefly for the initial request handling at 1. Everything after that is just IO, or app-level threads.

[Sente](https://www.taoensso.com/sente) uses this approach. It’d be a lot of code to look through, but the main relevant bits are probably [here](https://github.com/taoensso/sente/blob/102c5b0b2ca8e1ce931655f146d770287f4a1b8b/src/taoensso/sente/server_adapters/http_kit.clj#L26) and [here](https://github.com/taoensso/sente/blob/102c5b0b2ca8e1ce931655f146d770287f4a1b8b/src/taoensso/sente.cljc#L983).

# Approach 2

James Reeves has written something [here](https://github.com/ring-clojure/ring-examples/tree/master/websocket-chat) that could maybe be ported to http-kit (or at least linked to from http-kit's docs)?

# Assistance welcome!

PR very welcome if someone wanted to help out with this 🙏

The [Wiki files](https://github.com/http-kit/http-kit/tree/master/wiki) are part of the main http-kit repo.

jf:
I'd be happy to write up something for approach 1. Not able at this point in time to do approach 2 though, if that's ok.

ptaoussanis:
@jf Totally okay, thanks Jeffrey!

jf:
I have the PR ready @ptaoussanis , but will definitely need your input / help to really flesh things out. Thanks!

ptaoussanis:
Merged manually, thanks again Jeffrey! 🙏

PR #572, opened 2024-06-02
wiki/3-Server.md: add writeup for websockets (was -> "start writeup for WebSocket handling")

jf:
as promised in https://github.com/http-kit/http-kit/issues/570#issuecomment-2143496684

I wasn't too sure about the exact placement or structure, but figured I could just start, and you can then edit or shift as necessary. In that vein, I have only the barest of intros for the 2 approaches, and the main body of my work is in the demo code instead.

### DIFF
--- a/wiki/3-Server.md
+++ b/wiki/3-Server.md
@@ -106,7 +106,7 @@ On the surface, except for syntactical differences, these really look the same. 
 
 - with the unified API, you really only need to write one set of code (even right down to using the same `data` argument for `send!`) server-side if you want to support both WebSockets, and HTTP long-polling (perhaps as a fallback). Not so if you use the Ring WebSocket API. You will need to write separate code to handle HTTP long-polling.
 
-- with the unified API, detecting success or failure for `send!` (which is asynchronous already, by the way) is easy enough: just check the return value, no callbacks necessary. With the Ring WebSocket API, if you intend to send asynchronously, you will need to construct your own callback functions with each `send` (no data is otherwise provided to callbacks to link the success/failure back to the specific `send`).
+- with the unified API, detecting success or failure for `send!` (which is asynchronous already, by the way) is easy enough: just check the return value, no callbacks necessary. With the Ring WebSocket API, if you intend to send asynchronously, you will need to construct your own callback functions with each `send` (essentially closures; no data is otherwise provided to callbacks to link the success/failure back to the specific `send`).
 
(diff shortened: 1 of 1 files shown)


jf:
FYI I figured out the code for using WebSockets in Ring. I'm afraid this isn't the best code for showcasing / contrasting the 2 approaches though.

It looks like (at least for me, based on my reading) that one key difference is in how errors are handled when you send a message: for the unified API, you can handle it without callbacks, but for Ring, you'll need to handle it in a callback.

As mentioned at https://github.com/http-kit/http-kit/issues/570#issuecomment-2143868478, I will really need your help to really do the explanation and contrasting justice.

jf:
an update: I added the beginnings of some notes contrasting the 2 approaches.

A couple of notes:
1. I decided to refrain from saying "HTTP long-polling/streaming", but only "HTTP long-polling", because it seems that "HTTP streaming" can (is?) be associated with streaming media over HTTP, which is its own separate thing (especially if you get into [adaptive bitrate streaming](https://www.cloudflare.com/learning/video/what-is-adaptive-bitrate-streaming/)).
2. as usual, please vet the accuracy of my points.

Thank you.

ptaoussanis:
@jf Thanks Jeffrey! I'll take a closer look at this next time I'm on batched http-kit work.

Definitely feel free to keep things short/simple; something basic would already be a big upgrade on what we have now 👍

Really appreciate your help on this.

ptaoussanis:
Merged manually, thanks again Jeffrey! 🙏

jf:
thanks. Comparing before and after, I see how you've improved the code... (and even the formatting?)

One question: this is more out of curiosity, but in my PR, I do note that Ring's websocket API is experimental (as per https://github.com/ring-clojure/ring/wiki/WebSockets). Is this not the case?

ptaoussanis:
> Comparing before and after, I see how you've improved the code... (and even the formatting?)

Nothing significant, just some superficial changes to hopefully help keep things consistent with other code style in the wiki.

> One question: this is more out of curiosity, but in my PR, I do note that Ring's websocket API is experimental (as per https://github.com/ring-clojure/ring/wiki/WebSockets). Is this not the case?

I was under the impression that the API became stable at some point, though I may very well be mistaken! Ring is currently at [1.12.2](https://github.com/ring-clojure/ring/blob/master/CHANGELOG.md), though now that you mention it - the CHANGELOG doesn't make any specific mention of the API dropping its "experimental" status.

The current [protocol code](https://github.com/ring-clojure/ring/blob/master/ring-websocket-protocols/src/ring/websocket/protocols.clj) doesn't appear to be obviously marked as experimental, at least.

Would be happy to add the mention of "experimental" back in if I'm wrong! Thanks for checking on that.

Issue #574, opened 2024-09-16
[bug] java.util.zip.ZipException for deflate encoding on certain sites


pieterbreed:
Certain sites have a `deflate` encoding that does not carry Zip information. An example:

```clj
@(http/get "https://httpbin.org/deflate" {:as :text})
```

results in

```clj
{:opts {:as :text, :method :get, :url "https://httpbin.org/deflate"}, :error #error {
 :cause "invalid stored block lengths"
 :via
 [{:type java.util.zip.ZipException
   :message "invalid stored block lengths"
   :at [java.util.zip.InflaterInputStream read "InflaterInputStream.java" 164]}]
 :trace
 [[java.util.zip.InflaterInputStream read "InflaterInputStream.java" 164]
  [java.io.FilterInputStream read "FilterInputStream.java" 106]
  [org.httpkit.client.RespListener unzipBody "RespListener.java" 92]
  [org.httpkit.client.RespListener onCompleted "RespListener.java" 136]
  [org.httpkit.client.Request finish "Request.java" 70]
  [org.httpkit.client.HttpClient doRead "HttpClient.java" 243]
  [org.httpkit.client.HttpClient run "HttpClient.java" 519]
  [java.lang.Thread run "Thread.java" 840]]}}
```

ptaoussanis:
Have merged Pieter's fix manually - thanks Pieter! 👍

PR #575, opened 2024-09-16
check if content-encoding `deflate` carries zip envelope when setting up Inflater

pieterbreed:
fixes http-kit/http-kit#574

This fix was cribbed from https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/entity/DeflateInputStream.java

### DIFF
--- a/src/java/org/httpkit/client/RespListener.java
+++ b/src/java/org/httpkit/client/RespListener.java
@@ -10,6 +10,7 @@ import java.util.TreeMap;
 import java.util.concurrent.ExecutorService;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.Inflater;
+import java.util.zip.Deflater;
 import java.util.zip.InflaterInputStream;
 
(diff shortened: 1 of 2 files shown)


pieterbreed:
I ran the unit tests locally. The new test passed, resulting in this output 

```json
{
  "deflated": true, 
  "headers": {
    "Content-Length": "0", 
    "Host": "httpbin.org", 
    "User-Agent": "http-kit/2.0", 
    "X-Amzn-Trace-Id": "Root=1-66e8ab87-2090014b2ba1852134e59388"
  }, 
  "method": "GET", 
  "origin": "105.233.68.235"
}
```

I still have one failure, but I had it on `origin/master` too. `testDecodeChunkSplits(org.httpkit.client.HttpClientDecoderTest)`

ptaoussanis:
@pieterbreed Hi Pieter, thanks so much for identifying the problem here - and for your fix! 🙏🎉 Have merged manually 👍

Issue #375, opened 2018-05-12
Channel reuse can result in incorrect HTTP response


osbert:
This is related to https://github.com/ptaoussanis/sente/issues/260 but I have since finally taken the time to properly construct a test case and also am able to explain the issue much better.

When multiple HTTP requests are transmitted over a persistent connection, http-kit re-uses the same underlying channel between requests.  This causes issues when there is code that attempts to issue multiple delayed `send!` calls because the channel gets reset back to open during HTTP request processing for each succeeding request over the same socket.

In particular because of how sente is written, the logic to emulate a websocket using ajax is something like, when you have server events queue them up but if the channel is open, go ahead and send.  (It is not really clear to me if this should be regarded as a flaw in how sente attempts to emulate websockets.  It could be argued sente should not attempt to call `send!` on the same channel twice but ... I think there may not be a good way of determining that other than the current `open?` check).

Please see `org.httpkit.server-test/test-channel-reuse-async` in https://github.com/osbert/http-kit/blob/channel-reuse-test/test/org/httpkit/server_test.clj#L409 for a test that attempts to illustrate the potential issue.

I think one temporary workaround is to send back `Connection: close` in headers so that your HTTP consumers do not attempt to use a persistent connection which should prevent channel re-use between requests.

Still thinking about how best to fix, but figured I'd share my work so far.  Thanks, and please let me know if you have any thoughts or ideas you'd like to me explore.

osbert:
One suggested fix is the following patch. The intent is to change it so that a new channel is created for every request. Presumably there are some performance implications but those seemed secondary to correctness. The argument is effectively that even if `sente` is using the API incorrectly, we should be doing what we can prevent incorrect behavior.

```
modified   src/java/org/httpkit/server/HttpServer.java
@@ -150,9 +150,9 @@ public class HttpServer implements Runnable {
         try {
             boolean sentContinue = false;
             do {
-                AsyncChannel channel = atta.channel;
                 HttpRequest request = atta.decoder.decode(buffer);
                 if (request != null) {
+										AsyncChannel channel = new AsyncChannel(key, this);
                     channel.reset(request);
                     if (request.isWebSocket) {
                         key.attach(new WsAtta(channel, maxWs));
```

ptaoussanis:
Osbert's proposed fix has been merged, will be included in upcoming v2.7 beta.
Thanks again for the awesome work @osbert, it's much appreciated! 🚀

PR #579, opened 2024-10-10
Fix #578

### LINKED ISSUE #578: When using http-kit async channels to implement Server Sent Events the channel is never closed even when writing to an event-stream a client has closed.
When you use http-kit async channels to implement Server Sent Events, the channel is never closed even when writing to a channel that the client has disconnected from. `:on-close` callbacks are also never called. 

I had a look at the sente code, but it closes the connection on the sever side after each send when over http (i.e doesn't keep the event-stream open) so would not have run into this issue. 

Minimal repro steps (httpkit 2.8.0, java 21, clojure 1.12.0)

```clojure
(ns app.core
  (:require
   [org.httpkit.server :as hk]))

(defonce clients (atom #{}))

(defn format-event [body]
  (str "data: " body "\n\n"))

(defn send! [ch message]
  (hk/send! ch
    {:status 200
     :headers
     {"Content-Type"      "text/event-stream"
      "Cache-Control"     "no-cache, no-store"}
     :body   (format-event message)}
    false))

(defn handler-sse [req]
  (hk/as-channel req
    {:on-open  (fn [ch]   (swap! clients conj ch))
     :on-close (fn [ch _]
                 (swap! clients disj ch))}))


(defn broadcast-message-to-connected-clients! [message]
  (run! (fn [ch] (send! ch message)) @clients))

(def app
  (fn handler [{:keys [request-method uri] :as req}]
    (if (= [:get  "/"] [request-method uri])
      (handler-sse req)
      {:status 404})))

(comment
  (def server (hk/run-server #'app {:port 8080}))

  ;; stop server
  (server))

(comment
  ;; 1. Open a terminal and connect
  ;; curl localhost:8080 -vv

  ;; 2. Send some messages you should see them in your
  ;; terminal window
  (broadcast-message-to-connected-clients! "HELLO")

  ;; 3. You should see your client connected in here
  @clients

  ;; 4. Close the terminal
  
  ;; 5. Send a bunch of messages
  ;; in case there's buffering going on 
  (broadcast-message-to-connected-clients! "HELLO")

  ;; 6. Your client should be gone from the list as
  ;; the send should fail and close the channel
  @clients

  ;; 7. But instead the channels remain open

  ;; you can also run send over the connections
  ;; and they return true, which means message send successful.
  (mapv (fn [ch] (send! ch "Hello")) @clients)
  ;; but the message send should fail as the connection has been
  ;; closed by the client
  )
```
Let me know if you need more details.

andersmurphy:
The change from #375 was preventing the close handler from firing. This was due to creating new AsyncChannel which did not have a close handler set.

To fix this we only create a new channel if the previous one was closed. This prevents the issue of closed channels being reopened and reused described in #375. But, still ensures the close handler for a channel is called correctly. 

Added a comment in the same style as #375 as it was quite useful for understanding the intent of the change. 

I've tested this with the minimum repro described in issue #578 and it works. However, I am assuming my understanding of #375 is correct and the issue was specifically to do with re-opening closed connections. 

I'm going to do some more aggressive testing on the demo app where I ran into the problem originally, flood it with requests etc, see if anything weird happens.

### DIFF
--- a/src/java/org/httpkit/server/HttpServer.java
+++ b/src/java/org/httpkit/server/HttpServer.java
@@ -214,8 +214,20 @@ public class HttpServer implements Runnable {
                     // AsyncChannel channel = atta.channel;
                     //
                     // Ref. #375, thanks to @osbert for identifying this risk!
-
-                    AsyncChannel channel = new AsyncChannel(key, this); // This okay?
+                    //
+                    // The change from #375 was preventing the close handler from firing.
+                    // This was due to creating new AsyncChannel which did not have a
+                    // close handler set.
+                    //
+                    // To fix this we only create a new channel if the previous one was
+                    // closed. This prevents the issue of closed channels being reopened 
+                    // and reused described above. But, still ensures the close handler 
+                    // for a channel is called correctly.                    
+                    // AsyncChannel channel new AsyncChannel(key, this);
+                    //
+                    // Ref. #578
+                    //
+                    AsyncChannel channel = atta.channel.isClosed() ? new AsyncChannel(key, this) : atta.channel;
 
                     if (status.get() != Status.RUNNING) {
                         request.isKeepAlive = false;
--- a/test/org/httpkit/server_test.clj
+++ b/test/org/httpkit/server_test.clj
@@ -533,6 +533,51 @@
       (is (= "1" (:body (get @responses_ 1))))
       (is (= "2" (:body (get @responses_ 2)))))))
 
+(deftest test-channel-async-client-side-close  ; Ref. #578 #579
+  (http/with-async-connection-pool {:threads 1}
+    (let [ch_          (atom nil)
+          fired-close_ (atom nil)
+          captured_ (atom []) ; [<send-success?> ...]
+          capture!  (fn [result] (swap! captured_ conj result) result)
+          sse-event {:status 200
+                     :headers
+                     {"Content-Type"  "text/event-stream"
+                      "Cache-Control" "no-cache, no-store"}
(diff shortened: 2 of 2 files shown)


ptaoussanis:
@andersmurphy Thanks for this Anders!

Will try take a closer look this weekend. I'd like to try understand a little better what's actually going on underneath - and what the implications of this change might be.

Looks promising though!

If it's not very difficult to do, do you think it'd make sense to try add a simplified regression test that'd have caught the issue with your repro in #578?

andersmurphy:
Yeah definetly worth looking into further as I'm only confident this fixes one side of the problem (and it might not even be for the right reasons). 

I'll have a read through the linked sente issue that was in #375 to understand what the sente side of the issue was. Hopefully, it was just to do with dead connections being re-used and that messing with internal app state.

I'll see If I can cobble together some regression tests. Would be good to catch other things that could break long lived SSE connections. I noticed  PR #514 would most likely have broken them too.

___

As an aside, I was benchmarking*  jetty vs http-kit on our app the other day, httpkit is like 30-50% faster, it's really wild. I'd seen something close to 20% before but that was just with small responses, with larger responses http-kit pulls ahead even more. It's performance also doesn't seem to suffer as much as jetty on really high loads. 

*Usual benchmarking caveats apply.

andersmurphy:
That's a regression test added.

Also the good news is there's a regression test for #375 and this change doesn't break that test. 

Let me know if you want me to make any changes. Cheers.

ptaoussanis:
Merged manually

Issue #581, opened 2024-11-30
DynamicBytes has no protection against allocation of large arrays


reutsharabani:
https://github.com/http-kit/http-kit/blob/76b869fc34536ad0c43afa9a98d971a0fc32c644/src/java/org/httpkit/DynamicBytes.java#L16

This blows up with `after` exceeds max array size (Integer.MAX_VALUE).

I'm not sure what the effect of this is, or what the fix should be, but I ran into this when reading large responses. Had to fork the library, prevent it and throw an IOException downstream instead of letting it blow up.

Just thought I'd raise this for thought. I don't think my fix is the right thing to do but at least it let's us handle the error. Maybe there is something sensible you can think of.

To restore this just make the over-allocation factor 999% instead of 33% - should blow up faster :)

Thank you for your work!

ptaoussanis:
@reutsharabani Hi Reut, thanks for pinging about this! PR welcome for throwing a clear IOException here. Someone might have a better / alternative idea, but until then sounds like your change would anyway be an improvement.

PR #582, opened 2024-11-30
Draft: throw exception on arrays that are too big to allocate

reutsharabani:
This is just a sample of one way to mitigate [this](https://github.com/http-kit/http-kit/issues/581). I don't know if it's good.

I think maybe subclassing IOException instead of using a RuntimeException is best but then you have to propagate it everywhere.

### DIFF
--- a/src/java/org/httpkit/ArrayTooLargeException.java
+++ b/src/java/org/httpkit/ArrayTooLargeException.java
@@ -1,0 +1,9 @@
+package org.httpkit;
+
+import java.lang.RuntimeException;
+
+public class ArrayTooLargeException extends RuntimeException {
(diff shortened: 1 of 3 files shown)


src/java/org/httpkit/ArrayTooLargeException.java
```diff
@@ -0,0 +1,9 @@
+package org.httpkit;
+
+import java.lang.RuntimeException;
+
+public class ArrayTooLargeException extends RuntimeException {
```
reutsharabani, line 5: Maybe this should be an IOException. I'm not a java programmer and I don't know what's best here.

The a checked exception is a mess but may be worth it.

PR #584, opened 2025-01-10
Validate the port sent in the Host header

ianmuge:
Issue:
http-kit does not validate the port it attempts to extract from the Host header and as such a malformed port results in some errors thrown. This occurs when it is empty or when it is anything other than numeric. Examples:
```
curl -v localhost:3000/api/status -H "host: localhost:"
curl -v localhost:3000/api/status -H "host: localhost:()"
curl -v localhost:3000/api/status -H "host: localhost:es"
```
Resulting in:
```
java.lang.NumberFormatException: For input string: "es"
  at java.base/java. lang.NumberFormatException.forInputString(NumberFormatException.java:67)
  at java.base/java.lang.Integer.parseInt(Integer.java:662)
  at java.base/java.lang.Integer.value0f(Integer.java:989)
  at org.httpkit.server.HttpRequest.setHeaders(HttpRequest.java:83)
  at org.httpkit.server.HttpDecoder.readHeaders(HttpDecoder.java:269)
  at org.httpkit.server.HttpDecoder.decode(HttpDecoder.java:186)
  at org.httpkit.server.HttpServer.decodeHttp(HttpServer.java:173)
  at org.httpkit.server.HttpServer.doRead(HttpServer.java:258)
  at org.httpkit.server.HttpServer.run(HttpServer.java:378)
  at java.base/java. lang.Thread. run(Thread.java:1583)
```

Fix:
Validate that the port is not empty and if we fail to parse it ignore the error and continue decoding the request

### DIFF
--- a/src/java/org/httpkit/server/HttpRequest.java
+++ b/src/java/org/httpkit/server/HttpRequest.java
@@ -92,7 +92,13 @@ public class HttpRequest {
             int idx = h.lastIndexOf(':');
             if (idx != -1 && idx > ipv6end) {
                 this.serverName = h.substring(0, idx);
-                serverPort = Integer.valueOf(h.substring(idx + 1));
+                String serverPortCandidate = h.substring(idx + 1);
+                if (!serverPortCandidate.isEmpty()) {
+                    try {
+                        serverPort = Integer.parseInt(serverPortCandidate);
+                    } catch (Exception ignore) {
+                    }
+                }
             } else {
                 this.serverName = h;
             }


ptaoussanis:
@ianmuge Hi Ian, thanks for this! Seems broadly reasonable to me, but just to play devil's advocate - is it possible that some folks might prefer an error to be thrown?

In the case of an error- `serverPort` will be 80, right? Could that be misleading?

Any idea what other common servers tend to do when given an invalid port?

ianmuge:
Hello @ptaoussanis, 
I do not think the error would be much help here as the extraction of the port from this section of code complements the request and yes, the serverPort would still default to 80 if we can not successfully extract it according to [this](https://github.com/http-kit/http-kit/blob/1bfd1196e82fb162fd5eaa107d7a73de553b911b/src/java/org/httpkit/server/HttpRequest.java#L22), this is similar to how we treat the charset. 
In most typical scenarios where requests would be coming from a browser or integration platform, this would be set correctly and even in cases where the request passes through a proxy that modifies this header, it would still be sent out as a numeric.

ptaoussanis:
> I do not think the error would be much help here as the extraction of the port from this section of code complements the request

Sorry, could you maybe rephrase this? Not sure I understand.

> In most typical scenarios where requests would be coming from a browser or integration platform, this would be set correctly and even in cases where the request passes through a proxy that modifies this header, it would still be sent out as a numeric.

I'm not sure I'm following.

Currently (without this PR): if an invalid port header is provided, then the request will fail (throw), correct?

If this PR is merged and an invalid port header is provided, then the request will silently ignore the invalid header and [presume port 80](https://github.com/http-kit/http-kit/blob/1bfd1196e82fb162fd5eaa107d7a73de553b911b/src/java/org/httpkit/server/HttpRequest.java#L22), correct?

If those are both correct, then I'm confirming whether that's a behavioural change that would be universally desirable.

I'm not familiar with how other servers might handle this - maybe that'd be informative to check?

ianmuge:
> Currently (without this PR): if an invalid port header is provided, then the request will fail (throw), correct?

Yes, without the PR the request would throw an internal error but would not terminate the request, i.e the server will just keep spinning and consuming a thread.

> If this PR is merged and an invalid port header is provided, then the request will silently ignore the invalid header and [presume port 80](https://github.com/http-kit/http-kit/blob/1bfd1196e82fb162fd5eaa107d7a73de553b911b/src/java/org/httpkit/server/HttpRequest.java#L22), correct?

Correct

> If those are both correct, then I'm confirming whether that's a behavioural change that would be universally desirable.

Good point, I can make it throw a Protocol Exception as I have seen in other parts of the code when the requests fail validation checks
Looking at the RFC, [here](https://www.rfc-editor.org/rfc/rfc9110.html#name-host-and-authority) and [here](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23), I think you might be right that we should reject the request

ptaoussanis:
> without the PR the request would throw an internal error but would not terminate the request, i.e the server will just keep spinning and consuming a thread.

Ah, gotcha- thanks for clarifying that.

I was under the mistaken impression that the request would be rejected.

Is this actually a potential DoS vector then (assuming http-kit's not behind a proxy)?

> Good point, I can make it throw a Protocol Exception as I have seen in other parts of the code when the requests fail validation checks

That sounds reasonable to me, especially if the current behaviour is just to permanently and silently break.

> Looking at the RFC, [here](https://www.rfc-editor.org/rfc/rfc9110.html#name-host-and-authority) and [here](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23), I think you might be right that we should reject the request

Thanks for checking the RFCs. Would the idea be to return 400 in this case?

ianmuge:
- If the PR stays as is, we shall ignore the malformed port, assume port 80 and continue processing the request, 
- if we throw a ProtocolException we shall terminate the request and close the connection
- Without the PR we throw the error internally but the connection is never terminated

ianmuge:
> Is this actually a potential DoS vector then (assuming http-kit's not behind a proxy)?
 
Yes precisely

ptaoussanis:
Thanks for all the effort on this.

Apologies if I'm being really stupid, but can I just double check this assertion:

> without the PR the request would throw an internal error but would not terminate the request, i.e the server will just keep spinning and consuming a thread.

I've not had an opportunity to dig into the code properly, but it seems non-obvious to me that a `java.lang.NumberFormatException` throwing would keep a thread spinning and not terminate a request, but throwing a `ProtocolException` instead would appropriately terminate the request?

I may be missing something obvious, it's been a long day and I'm quite tired ^^

What's causing a DoS scenario in the first case?

ianmuge:
> I've not had an opportunity to dig into the code properly, but it seems non-obvious to me that a java.lang.NumberFormatException throwing would keep a thread spinning and not terminate a request, but throwing a ProtocolException instead would appropriately terminate the request? Is something higher up specifically looking for ProtocolException?

I am sorry for being unclear, The problem is that the NumberFormatException does not bubble up but with the ProtocolException being thrown it bubbles up and gets caught [here](https://github.com/http-kit/http-kit/blob/76b869fc34536ad0c43afa9a98d971a0fc32c644/src/java/org/httpkit/server/HttpServer.java#L248)  resulting in the connection getting closed

ptaoussanis:
> but with the ProtocolException being thrown it bubbles up and gets caught [here](https://github.com/http-kit/http-kit/blob/76b869fc34536ad0c43afa9a98d971a0fc32c644/src/java/org/httpkit/server/HttpServer.java#L248) resulting in the connection getting closed

Excellent, thanks for helping me understand 👍

Would it'd be useful to add a small unit test for the new behaviour while we're here?

ianmuge:
Sure thing, if you can guide me to somewhere in the tests that's slightly similar to this I can extend them

ptaoussanis:
Possibly somewhere approximately around [here](https://github.com/http-kit/http-kit/blob/76b869fc34536ad0c43afa9a98d971a0fc32c644/test/org/httpkit/server_test.clj#L251)?

Most of these basic tests just take the form of an http-kit client request against a running http-kit server to check the response.

ianmuge:
I have added some tests but I am not sure about the structure of the last ones but I hope they are suitable

ptaoussanis:
Fix has been merged manually, with some [minor tweaks](https://github.com/http-kit/http-kit/commit/e8ca029dc543fa386f01e155eff6eefa7bab544b).

Thanks again for your work on this Ian!

I've also opened a [related issue](https://github.com/http-kit/http-kit/issues/585) re: possible follow-up improvements. Would of course be happy if you also wanted to take a look at that.

ianmuge:
Thanks again for that

jumarko:
I just found this - very timely!

We are seeing the very same problem where malicious requests containing invalid port (such as "aaaa") in the Host header 
cause http-kit to hung.
As OP, I can see that such curl requests don't complete (at least not within a couple of minutes)
```
curl -H 'Host: localhost:aaaa' -I http://localhost:3003/
```

What's worse, 4 concurrent requests make the server/app totally unresponsive to new requests
```
for N in {1..4}; do curl -H 'Host: localhost:aaaa' -I http://localhost:3003/ &; done
```
- not sure what's so magical about 4 but it keeps accepting new requests after issuing only 3 concurrent requests (we have `:thread 8` in the options)


NOTE: This is running behind nginx, but there's nothing in the default nginx configuration that prevents such malformed Host headers (here's [one way](https://blog.shalvah.me/posts/fixing-the-host-header-vulnerability-with-nginx) to make nginx reject such requests).


Am I correct assuming that this PR will solve all these issues? 
When approximately could a new stable version containing this fix be released? 

Thanks!

ptaoussanis:
@jumarko Hi Juraj,

> not sure what's so magical about 4 but it keeps accepting new requests after issuing only 3 concurrent requests (we have :thread 8 in the options)

My guess also would have been a low thread count since old versions of http-kit used to default to 4 threads for the server. `:thread` should still work iirc, but note that it'd be better to set an explicit [`:worker-pool`](https://github.com/http-kit/http-kit/blob/de33aedd4afda7b66d86cd1726e3ce0bed2306db/src/org/httpkit/server.clj#L79). That'd also help here since it'd let you more easily inspect the pool to see what's going on in your case.

>  there's nothing in the default nginx configuration that prevents such malformed Host headers

Are you sure about this? If you're using nginx as a remote proxy, and matching based on `server_name` (domain) and `listen` (port) - I'd expect that to not even proxy the request for a bad port. I'm not sure how it'd match in that case, unless you're using a catch-all config?

I might be missing something though.

> Am I correct assuming that this PR will solve all these issues?

Depends what you mean by "all these issues". [`[http-kit "2.9.0-alpha2"]`](https://clojars.org/http-kit/versions/2.9.0-alpha2) should include the current PR at least 👍

jumarko:
Thanks for the fast reply!
I don't have access to the full nginx config (it's one of our clients) but it seems that it proxies it just fine.

I'll give the latest alpha a shot - thanks!

ptaoussanis:
You're welcome, please let me know if you still have any issues with 2.9.0-alpha2 👍

jumarko:
2.9.0-alpha2 seems to work fine

ptaoussanis:
Excellent, thanks for the confirmation 👍

PR #590, opened 2025-03-19
Specifying number of workers in NioClientSocketChannelFactory

### LINKED ISSUE #576: [tests] Fix Netty warnings in tests
For some time the unit tests (org.httpkit.ws-test?) print a warning:

> org.jboss.netty.channel.socket.nio.AbstractNioWorkerPool
WARNING: Failed to get all worker threads ready within 10 second(s). Make sure to specify the executor which has more threads than the requested workerCount. If unsure, use Executors.newCachedThreadPool().

Seems to also slow down the tests so would be nice to get this resolved - might be a good first PR if someone felt like taking a look 🙏

kolstae:
The default is double the available cpu cores

This seems to fix #576

I ran the tests using `lein test` an _not_ `rake` - hopefully that doesn't change anything


ptaoussanis:
Thanks Espen! Cheers :-)

PR #593, opened 2025-04-18
Make RespListener immediately stream response

### LINKED ISSUE #591: Client buffers entire stream contents before making them available to the handler
There are some issues about this topic for the server, but I think this is the first one for the client. See [Slack](https://clojurians.slack.com/archives/C053AK3F9/p1743546370792609) for a discussion

# Issue Description
When making a request to a streaming endpoint using http kit client (e.g., OpenAI chat completions or responses), the handler does not execute as soon as the headers are available, but rather as soon as all data is finished streaming from the source. This makes streaming non-functional when using http kit.

# Expected behavior
The handler should be able to process chunks as soon as they're available

# Example code

```clojure
(comment (try (http-kit/post "https://api.openai.com/v1/responses"
                             {:as :stream
                              :headers {"Authorization" (str "Bearer " env/openai-token)
                                        "Content-Type" "application/json"}
                              :body (cheshire/generate-string
                                     {:model "gpt-4o"
                                      :input "Write a 2000 word short story about a cat with a hat"
                                      :stream true})}
                             (fn [{:keys [body]}]
                               (println "Got body")
                               (let [reader (io/reader body)]
                                 (doseq [line (line-seq reader)]
                                   (println line)))))

              (catch Exception e
                (print (ex-data e)))))
```

In the above minimal example (2000 words is chosen to test the performance on a substantial output), the print statement "Got body" takes about 20-40 seconds to execute. Immediately after, all the chunks are rapidly printed to the screen. This indicates that the handler did not execute until the entire stream was done producing.

In contrast, this minimal example using Aleph HTTP:

```
(comment (let [time (System/nanoTime)
               resp @(aleph/post "https://api.openai.com/v1/responses"
                                 {:headers {"Authorization" (str "Bearer " env/openai-token)}
                                  :raw-stream? true
                                  :body (cheshire/generate-string
                                         {:model "gpt-4o"
                                          :input "Write a 2000 word short story about a cat with a hat"
                                          :stream true})})]
           (println (str "Got response after " (- (System/nanoTime) time)))
           (s/consume
            (fn [chunk]
              (-> chunk
                  bs/to-string
                  println))
            (:body resp))))
```

The first chunk executes in less than 1s after making the rest, and the chunks print as soon as they are produced, which is the expected behavior.

# Impact
HTTP kit currently cannot be used as a client for use cases (e.g., chat bots) in which streaming is necessary for good user experience.

GAumala:
Fixes #591 

### How to test

At the moment I'm not sure how to write a test for this, but you can use this Node.js script to start a server that for every request streams the words 'one', 'two', 'three'... 'hundred'.

```javascript
const http = require('http');

const server = http.createServer((req, res) => {
  // Set headers for SSE (Server-Sent Events) or plain text streaming
  res.writeHead(200, {
    'Content-Type': 'text/plain',
    'Transfer-Encoding': 'chunked',
    'Cache-Control': 'no-cache',
    'Connection': 'keep-alive'
  });

  // Function to send numbers with delay
  async function streamNumbers() {
    for (let i = 1; i <= 100; i++) {
      const word = numberToWord(i);
      // Write the word followed by a comma and space (except the last one)
      res.write(`${word}${i < 100 ? ', ' : ''}`);
      
      // Add a delay to simulate processing time (250ms)
      await new Promise(resolve => setTimeout(resolve, 250));
    }
    
    // End the response when done
    res.end();
  }

  // Start streaming
  streamNumbers().catch(err => {
    console.error('Stream error:', err);
    res.end();
  });
});

// Helper function to convert numbers to words
function numberToWord(num) {
  const units = ['', 'one', 'two', 'three', 'four', 'five', 'six', 'seven', 'eight', 'nine'];
  const teens = ['ten', 'eleven', 'twelve', 'thirteen', 'fourteen', 'fifteen', 'sixteen', 
                'seventeen', 'eighteen', 'nineteen'];
  const tens = ['', '', 'twenty', 'thirty', 'forty', 'fifty', 'sixty', 'seventy', 
               'eighty', 'ninety'];

  if (num === 0) return 'zero';
  if (num < 10) return units[num];
  if (num < 20) return teens[num - 10];
  if (num < 100) {
    const ten = Math.floor(num / 10);
    const unit = num % 10;
    return tens[ten] + (unit ? '-' + units[unit] : '');
  }
  return num.toString(); // For 100+ just return the number
}

const PORT = 3000;
server.listen(PORT, () => {
  console.log(`Server running at http://localhost:${PORT}`);
});
```

Then on a REPL use this namespace to verify that the chunks get received one by one:

``` clojure
(ns stream-responses
  (:require [org.httpkit.client :as hk]))

(defn print-streaming-bytes
  "Reads bytes from an InputStream and prints them as they arrive."
  [input-stream]
  (let [buffer-size 256
        encoding "UTF-8"
        buffer (byte-array buffer-size)]
    (try
      (loop []
        (let [bytes-read (.read input-stream buffer)]
          (when (pos? bytes-read)
            (let [chunk (String. buffer 0 bytes-read encoding)]
              (print chunk)
              (flush)
              (recur)))))
      (catch Exception e
        (println "Error reading stream:" (.getMessage e)))
      (finally
        (.close input-stream)))))

(defn run [url]
  (let [res @(hk/get url {:as :stream})]
    (print-streaming-bytes (:body res))))
```

Usage:

```clojure
(require 'stream-responses)
(in-ns 'stream-responses)
(run "http://localhost:3000")
;; => one,
;; => two, three, four,
;; =>five,
;; => ...
```

### DIFF
--- a/src/java/org/httpkit/client/Decoder.java
+++ b/src/java/org/httpkit/client/Decoder.java
@@ -160,6 +160,27 @@ public class Decoder {
         }
     }
 
+    private int parseContentLength() {
+        String cl = HttpUtils.getStringValue(headers, CONTENT_LENGTH);
+        if (cl == null) {
+            // response does not have content length
+            // it either has no body or is variable length
+            return -1;
+        }
+
+        try {
+          return Integer.parseInt(cl);
+        } catch (NumberFormatException ex) {
+          long parsed = Long.parseLong(cl);
+          if (parsed > Integer.MAX_VALUE) {
+            // Content is is larger that Integer.MAX_VALUE
+            // Let's pretend it has variable length.
+            return -1;
+          }
+          throw ex;
+        }
+    }
+
     private void readHeaders(ByteBuffer buffer) throws LineTooLargeException, AbortException, ProtocolException {
         String line = lineReader.readLine(buffer);
         while (line != null && !line.isEmpty()) {
@@ -180,14 +201,15 @@ public class Decoder {
             if (headers.containsKey(TRAILER))
               chunkTrailerExpected = true;
         } else {
-            String cl = HttpUtils.getStringValue(headers, CONTENT_LENGTH);
-            if (cl != null) {
-                readRemaining = Integer.parseInt(cl);
+            int cl = parseContentLength();
+            if (cl >= 0) {
+                readRemaining = cl;
                 if (readRemaining == 0) {
                     state = ALL_READ;
                 } else {
                     state = READ_FIXED_LENGTH_CONTENT;
                 }
+
             } else if (emptyBodyExpected) {
                 state = ALL_READ;
             } else {
--- a/src/java/org/httpkit/client/RespListener.java
+++ b/src/java/org/httpkit/client/RespListener.java
@@ -4,6 +4,9 @@ import org.httpkit.*;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
 import java.nio.charset.Charset;
 import java.util.Map;
 import java.util.TreeMap;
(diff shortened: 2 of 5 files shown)


ptaoussanis:
@GAumala This looks promising, thanks a lot Gabriel! 🙏

I'm just focused on some other batched work atm, so hopefully someone else can help review. If not, I'll try come back to this myself w/in a couple weeks 👍

GAumala:
I wanted to see if these branch could also handle downloading large files (~4 GB) using `InputStream`, so I did some testing. 

If the server uses uses chunked transfer encoding it works, but if the response's content length is a fixed number bigger than `Integer.MAX_VALUE` it crashes. I added a commit to handle this.

ptaoussanis:
> Looking at the patch I wonder if it would make sense for [this](https://github.com/http-kit/http-kit/pull/593/files#diff-ad3d3770bcf534a552b471e340c23c5f0eec0dc9d1b7370ffed691188149ebafR138) 8kb buffer size on the piped input stream to be configurable? I'm not confident that it's performance-sensitive, but it feels like it might be.

Ideas welcome for how we might make this configurable, but would recommend that someone first measure to confirm if it'd really be worth the added cost+complexity.

ptaoussanis:
**Update**: https://github.com/http-kit/http-kit/commit/cdb06f2631022086dc363be8677d46b719d371a2 has just been reverted due to an unexpected [regression](https://github.com/http-kit/http-kit/issues/602).

See the following call:

```
(-> (org.httpkit.client/get "https://sleepingpill.javazone.no/public/allSessions/javazone_2025" {:as :stream}) deref :body slurp)
```

- http-kit client 2.8.1        returns a JSON    string (correct)
- http-kit client 2.9.0-alpha2 returns a garbled string (incorrect)

What's going on?

- The URI above is gzip compressed
- 2.8.1        does    auto gzip decompression on the response body
- 2.9.0-alpha2 does NO auto gzip decompression on the response body

New/updated PR welcome to reintroduce https://github.com/http-kit/http-kit/pull/593 with:

1. Restored                auto unzip support
2. Added test/s to confirm auto unzip support

Further discussion at https://clojurians.slack.com/archives/C01ALH5JGRJ/p1757409736717179

ptaoussanis:
Going to assume this is resolved, closing as part of issue triage.

PR #594, opened 2025-05-01
Remove default Content-Length: 0 for bodyless requests

### LINKED ISSUE #583: http-kit sends Content-Length: 0 in GET requests
One of the APIs I'm trying to connect to has started to deny requests when using http-kit.

I narrowed it down to the following:
 - The API is using CloudFront in Strict Mode, which does not allow Content-Length headers in GET requests [0] because they are a "SHOULD NOT" in the HTTP/1.1 RFC [1]
 - http-kit always sends a content-length header in all requests, including GET requests [2]

My current workaround is to use clj-http, which does not send the Content-Length header in GET requests.

I believe http-kit should not send any Content-Length header in GET or HEAD requests, to more strictly comply with the RFC.

[0] https://stackoverflow.com/questions/77288320/aws-application-load-balancer-rejecting-requests-with-content-length-0#

[1] [HTTP/1.1 RFC7230](https://www.rfc-editor.org/rfc/rfc7230#section-3.3.2)
"A user agent SHOULD NOT send a Content-Length header field when the request message does not contain a payload body and the method semantics do not anticipate such a body."

[2] https://github.com/http-kit/http-kit/blob/3831982/src/java/org/httpkit/client/HttpClient.java#L416

GAumala:
Fixes #583 

I could not find any good reason for the client to send `Content-Length: 0` in a bodyless request. This behavior is unnecessary and non-compliant with the HTTP/1.1 best practices: 

- RFC 7230 (Section 3.3.2) states that `Content-Length` should only be used when a message body is present.
- Bodyless methods like GET or HEAD should omit `Content-Length` entirely, as its absence implies no body.
- Some servers or proxies may misinterpret `Content-Length: 0` as an empty body rather than no body, leading to edge-case issues.

 Users who explicitly need this header (e.g., for legacy server compatibility) can still add it manually.

### DIFF
--- a/src/java/org/httpkit/client/HttpClient.java
+++ b/src/java/org/httpkit/client/HttpClient.java
@@ -411,10 +411,10 @@ public class HttpClient implements Runnable {
         ByteBuffer bodyBuffer = HttpUtils.bodyBuffer(body);
 
         if (body != null) {
-            headers.putOrReplace("Content-Length", Integer.toString(bodyBuffer.remaining()));
-        } else {
-            headers.putOrReplace("Content-Length", "0");
+            String value = Integer.toString(bodyBuffer.remaining());
+            headers.putOrReplace("Content-Length", value);
         }
+
         DynamicBytes bytes = new DynamicBytes(196);
         bytes.append(method.toString()).append(SP).append(path);
         bytes.append(" HTTP/1.1\r\n");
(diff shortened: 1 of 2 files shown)


ptaoussanis:
@GAumala Great, thanks for the fix and clear PR Gabriel! Cheers :-)

> Users who explicitly need this header (e.g., for legacy server compatibility) can still add it manually.

Will make a note of this in the next release notes 👍

In the meantime have pushed [v2.9.0-alpha6](https://clojars.org/http-kit/versions/2.9.0-alpha6) based on master with this commit.

PR #596, opened 2025-07-04
Fix #366 - httpclient get, had to force {:as :stream} since otherwise .xlsx was corrupted

### LINKED ISSUE #366: httpclient get, had to force {:as :stream} since otherwise .xlsx was corrupted
Implemented a reverse proxy using ring + httpkit client, which transforms the .html pages. (Mostly removes Wordpress yuck).

Other pages like images, .xlsx files etc, should just pass through. Just taking the :body from http/get and give it to the ring handler works fine, but NOT for .xlsx. The are returned as String, and somehow the content is corrupted in the process. I solved it like this, but is this really the way to do it? Should they even been converted into String from the start?

```
  ;; KLUDGE: Otherwise, httpkit will return these as String, and then they delivered corrupted
  (let [options (if (str/ends-with? url ".xlsx")
                  {:as :stream}
                  {})
        result @(http/get (make-origin url) options)

```

TheKostins:
Resolved #366 this via dedicated set of types that should not be treated as text. 
I've also discovered some other file formats that contain text or xml substring and added them to the set.

Should I also add tests to check this behavior?

This may be only my preference but I suggest to refactor the `isText()` to reduce nesting.
<details>

<summary>suggested `isText()` code</summary>

```java
 private boolean isText() {
        if (status.getCode() != 200) {
            return true;
        } // non 200: treat as text
        String type = HttpUtils.getStringValue(headers, CONTENT_TYPE);

        if (type == null) {
            return false;
        }

        type = type.toLowerCase();

        // TODO may miss something
        if (NON_TEXT_CONTENT_TYPES.contains(type)) {
            return false;
        } else {
            return type.contains("text") || type.contains("json") || type.contains("xml");
        }
    }
```
</details>

### DIFF
--- a/src/java/org/httpkit/HttpUtils.java
+++ b/src/java/org/httpkit/HttpUtils.java
@@ -67,6 +67,22 @@ public class HttpUtils {
     public static final String CONTENT_TYPE = "content-type";
     public static final String CONTENT_LENGTH = "content-length";
 
+    public static final Set<String> NON_TEXT_CONTENT_TYPES = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
+            "application/vnd.openxmlformats-officedocument.wordprocessingml.document",   // .docx
+            "application/vnd.openxmlformats-officedocument.wordprocessingml.template",   // .dotx
+            "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",         // .xlsx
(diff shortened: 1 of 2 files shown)


ptaoussanis:
@TheKostins Hi Konstantin, thanks for this! Apologies for the delay replying, merging manually now 👍

PR #601, opened 2025-08-29
avoid int overflow in DynamicBytes

reutsharabani:
Avoid overflow in DynamicBytes

- Promote idx to long before addition to prevent int overflow when computing (idx + more) * 1.33
- Keep double before allocating the array to make sure we didn't overflow

### DIFF
--- a/src/java/org/httpkit/DynamicBytes.java
+++ b/src/java/org/httpkit/DynamicBytes.java
@@ -14,7 +14,8 @@ public class DynamicBytes {
 
     private void expandIfNeeded(int more) {
         if (idx + more > data.length) {
-            long after =  (long)((idx + more) * 1.33);
+            double after =  (((long)idx + more) * 1.33);
+            // java arrays are indexed by `int`
             if (after >= Integer.MAX_VALUE) {
                 throw new ContentTooLargeException("Cannot expand DynamicBytes array: requested size (" + after + ") exceeds java limits");
             }


Ramblurr:
@reutsharabani 

You seem to have found a subtle bug, many thanks for the PR. I am curious, did you trigger this bug in real life?

Having recently encountered the max array length in the JVM, I suggest that we change the comparison `if (after >= Integer.MAX_VALUE) ` to `if (after >= (Integer.MAX_VALUE - 8))`. We can put `Integer.MAX_VALUE - 8` in a static const field. 

If you don't want to change the PR, no worries, I can add that in a merge.

For more context see: 

```
    /**
     * A soft maximum array length imposed by array growth computations.
     * Some JVMs (such as HotSpot) have an implementation limit that will cause
     *
     *     OutOfMemoryError("Requested array size exceeds VM limit")
     *
     * to be thrown if a request is made to allocate an array of some length near
     * Integer.MAX_VALUE, even if there is sufficient heap available. The actual
     * limit might depend on some JVM implementation-specific characteristics such
     * as the object header size. The soft maximum value is chosen conservatively so
     * as to be smaller than any implementation limit that is likely to be encountered.
     */
    public static final int SOFT_MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 8;
```

source: https://github.com/openjdk/jdk/blob/d5ce66698d2f15c5f8316110a6118a10baa4013d/src/java.base/share/classes/jdk/internal/util/ArraysSupport.java#L842-L854

@ptaoussanis any other feedback?

src/java/org/httpkit/DynamicBytes.java
```diff
@@ -14,7 +14,8 @@ public DynamicBytes(int size) {
 
     private void expandIfNeeded(int more) {
         if (idx + more > data.length) {
-            long after =  (long)((idx + more) * 1.33);
+            double after =  (((long)idx + more) * 1.33);
+            // java arrays are indexed by `int`
             if (after >= Integer.MAX_VALUE) {
```
reutsharabani:
```suggestion
            // https://en.wikipedia.org/wiki/Criticism_of_Java#Large_arrays
            if (after >= (Integer.MAX_VALUE - 8)) {
```

PR #610, opened 2026-01-08
[fix] [#609] make-client use the default-ssl-configurer, if available

souenzzo:
Now `(make-client {})` is always equivalent to `default-client`

<details>
<summary> `ssl-configurer -> default-ssl-configurer` is a safe rename because it is private</summary>

I can't find any usage of `org.httpkit.client/ssl-configurer` in other libraries

https://grep.app/search?f.lang=Clojure&q=ssl-configurer
</details>

Edge case:
- if `default-ssl-configurer` is `nil` and you call `(make-client {})`, it will use `HttpClient$SSLEngineURIConfigurer/NOP`
- if `default-ssl-configurer` is `nil` and you call `(make-client {:ssl-configurer my-ssl})`, it will use `my-ssl`

### DIFF
--- a/src/org/httpkit/client.clj
+++ b/src/org/httpkit/client.clj
@@ -139,6 +139,13 @@
   (^SSLEngine [               ] (make-ssl-engine (SSLContext/getDefault)))
   (^SSLEngine [^SSLContext ctx] (.createSSLEngine ctx)))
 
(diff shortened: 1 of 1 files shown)


souenzzo:
Extra node: seems that `default-ssl-configurer`  will never be `nil`.

Issue #607, opened 2025-12-15
Client SNI compatibility issue


jefimm:
It looks like  HttpsRequest::doHandshake does not work correctly with nginx if the server is configured like this

```
server {
    listen 43 ssl default_server;
    listen [::]:443 ssl default_server;
    ssl_reject_handshake on;
}
```
setting ssl_reject_handshake off;  mitigates the issue however, the https://github.com/nginx/nginx-gateway-fabric does not allow to configure it, meaning all new servers based on https://github.com/nginx/nginx-gateway-fabric will not allow the client to connect
This is the stack trace observed with error
 javax.net.ssl.SSLHandshakeException: (unrecognized_name) Received fatal alert: unrecognized_name

```
sun.security.ssl.Alert.createSSLException (Alert.java:117)
sun.security.ssl.TransportContext.fatal (TransportContext.java:370)
sun.security.ssl.Alert$AlertConsumer.consume (Alert.java:287)
sun.security.ssl.TransportContext.dispatch (TransportContext.java:209)
sun.security.ssl.SSLTransport.decode (SSLTransport.java:172)
sun.security.ssl.SSLEngineImpl.decode (SSLEngineImpl.java:736)
sun.security.ssl.SSLEngineImpl.readRecord (SSLEngineImpl.java:691)
sun.security.ssl.SSLEngineImpl.unwrap (SSLEngineImpl.java:506)
sun.security.ssl.SSLEngineImpl.unwrap (SSLEngineImpl.java:482)
javax.net.ssl.SSLEngine.unwrap (SSLEngine.java:679)
org.httpkit.client.HttpsRequest.doHandshake (HttpsRequest.java:100)
org.httpkit.client.HttpClient.doRead (HttpClient.java:218)
org.httpkit.client.HttpClient.run (HttpClient.java:519)
java.lang.Thread.run (Thread.java:1583)
```

souenzzo:
@jefimm this issue is not fixed.
HTTP Kit is still unable to fetch SNI-enabled servers when using `:insecure? true`.

Ramblurr:
I've dug into this further and the CI failures on PR #608 reveal more.. The issue isn't just about SNI itself - it's that the SNI configurer also enables hostname verification as a side effect:

https://github.com/http-kit/http-kit/blob/22c615f03f8c3e7c4d5d0313745b4a6e5f285300/src/org/httpkit/sni_client.clj#L41

[...]

PR #611, opened 2026-01-08
[fix] [client] [#607] Enable SNI with :insecure? true

Ramblurr:
This is an alternative to pr #608, which has what I think is the correct fix for #607.

Thanks to Enzzo for taking the first stab!
I don't want to steal anyone's glory so feel free to wrap this into their commit and give them the credit.

### DIFF
--- a/src/java/org/httpkit/server/HttpServer.java
+++ b/src/java/org/httpkit/server/HttpServer.java
@@ -251,7 +251,8 @@ public class HttpServer implements Runnable {
                 }
             } while (buffer.hasRemaining()); // consume all
         } catch (ProtocolException e) {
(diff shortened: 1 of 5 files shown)


souenzzo (approved):
Nice tests btw :)

souenzzo (changes requested):

ptaoussanis:
@Ramblurr @souenzzo This looks great- thank you Casey, Enzzo!

Will try get this merged later today 👍

souenzzo (approved):

ptaoussanis:
Apologies for the long delay, have been relocating (among other distractions 😅)- merged manually now! 👍

PR #616, opened 2026-07-05
Add support for HTTP QUERY method

Nozistance:
https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body

### DIFF
--- a/src/java/org/httpkit/HttpMethod.java
+++ b/src/java/org/httpkit/HttpMethod.java
@@ -11,7 +11,8 @@ public enum HttpMethod {
     CONNECT(intern("connect")), PATCH(intern("patch")), PROPFIND(intern("propfind")),
     PROPPATCH(intern("proppatch")), LOCK(intern("lock")), UNLOCK(intern("unlock")),
     REPORT(intern("report")), ACL(intern("acl")), MOVE(intern("move")),
-    COPY(intern("copy")), MKCOL(intern("mkcol")), LIST(intern("list"));
+    COPY(intern("copy")), MKCOL(intern("mkcol")), LIST(intern("list")),
+    QUERY(intern("query"));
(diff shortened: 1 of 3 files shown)


Ramblurr (changes requested):
@Nozistance thanks for this PR.

There's an oversight.. the existing redirect rewrite rules in the client are incorrect:  `src/org/httpkit/client.clj:339-350`

RFC 10008 says 301, 302, 307, and 308 redirects for QUERY should be followed with a similar QUERY request to the new target. It explicitly says the POST-to-GET exceptions for 301 and 302 do not apply to QUERY.

Your patch therefore makes `hkc/query`  silently become GET on 301 and 302 redirects.

This is observable with the existing test route: a QUERY request to `/redirect?code=301`  or  `/redirect?code=302`  returns final body  `get` , not `query`.

Nozistance:
Yep, missed that, fixed that (6a97ce9)

Ramblurr (changes requested):
There is still an edge case when it comes to a 303 redirect on the client..

RFC 10008 section 2.5 treats a 303 differently from 301, 302, 307, and 308. Those other redirects should repeat the QUERY request at the new URI, but a 303 should follow the Location with a normal GET.. and the original QUERY content should not be sent again.

The updated code changes the method to GET, but it builds the redirected request from the original options and clears only `:query-params` and `:form-params`. If the caller supplied the query content using `:body` and a Content-Type header, both are retained, so the client sends a GET containing the original QUERY body.

The relevant redirect logic is at src/org/httpkit/client.clj:344-352

The new 303 test uses a bodyless request, so it does not catch this.

Could you clear the body and its content headers when following a QUERY 303, and add a test using a QUERY request with a body?

Issue #615, opened 2026-06-19
HTTP-Kit should join headers using `,`, folloing ring spec


souenzzo:
ring spec says that multiple headers should be joined with `,`

https://github.com/ring-clojure/ring/blob/master/SPEC.md#headers

Discovered by:

https://github.com/souenzzo/ring-adapter-test/pull/2#issuecomment-4747504698

Related: #501, #502, #108

ptaoussanis:
@souenzzo Hi Enzzo- thanks for identifying this, and for the fix (and related issues!).

@g23 Just a heads-up that it looks like the change I merged in #502 actually violates the Ring spec and so may need to be re-examined. If reverted, that'd unfortunately be a breaking change for you and anyone else taking advantage of the same behaviour.

@souenzzo Do you happen to have any idea what the motivation is for that Ring spec detail? 

From what I can tell the motivation for #501 seems broadly reasonable?

Thoughts very welcome!

souenzzo:
The `,` was introduced on this commit. I couldn't find any related discussions.

I agree that it's not the best idea, mainly because `,` is a valid header character/value. 

https://github.com/ring-clojure/ring/commit/05a34953f0a0c2ebed74b88eb287bc1a495ae34c

I'll investigate a bit more - and maybe open the discussion in the ring?

souenzzo:
RFC 7230 defines:

- https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2

```
A recipient MAY combine multiple header fields with the same field
   name into one "field-name: field-value" pair, without changing the
   semantics of the message, by appending each subsequent field value to
   the combined field value in order, separated by a comma. 
``` 

In `nodejs` (v26), if you try to send two headers with the same name, it will join both using `,`

```javascript
await fetch("http://localhost:1337", {headers: [["a", "1"], ["a", "2"]]})
```

results in:

```
GET / HTTP/1.1
host: localhost:1337
connection: keep-alive
a: 1, 2
``` 

Also, if you send multiple headers into a nodejs HTTP server, it will join the values values into a single `{a: "1, 2"}` header representation.

Given this context, I think that the ring spec is good enough and consistent with the specifications, and we should follow it.

ptaoussanis:
@souenzzo Thanks Enzzo, let's back up a bit / untangle things.

From what I can tell (please correct where wrong!):

[...]

PR #617, opened 2026-08-01
[server] WebSocket permessage-deflate (RFC 7692)

whilo:
Adds the `permessage-deflate` WebSocket extension to the server, with context
takeover.

Off unless the client offers it — a connection that does not negotiate the
extension takes the same code path it did before, so this should be inert for
existing users.

## Why context takeover

Compressing each message independently saves ~10% on small messages and is
barely worth the CPU. Context takeover — message N compressed against messages
1..N-1 — is what makes the extension pay for a stream of many similar messages.
On a sample of 500 small JSON-ish messages: **88 624 → 10 783 bytes (-88%)**.

That is also the motivation: a WebSocket carrying a stream of structurally
similar messages (RPC, replication, telemetry) is exactly the shape this helps,
and browsers offer the extension automatically, so it costs the application
nothing to benefit.

## Public surface

Two dynamic vars in `org.httpkit.server`:

- `*websocket-compression?*` (default true) — bind to false to refuse the
  offer. Worth doing when payloads are already compressed (images, video),
  where deflate spends CPU to make them slightly larger, or when per-connection
  memory matters more than bandwidth: context takeover keeps a deflate and an
  inflate window alive for the life of each connection.
- `*websocket-max-message-size*` — bounds a message's size **after**
  decompression. This is deliberately separate from `:max-ws`, which bounds the
  bytes actually received and therefore says nothing about the size after
  inflation; a small compressed frame can expand enormously. Exceeding it
  closes with 1009, the same status `WSDecoder` already uses.

## What is not implemented

`server_max_window_bits`. `java.util.zip` does not expose zlib's windowBits, so
the server cannot honour a smaller window and does not claim to — the parameter
is simply not echoed, which per RFC 7692 §7.1.2.2 means a 15-bit window.
`client_max_window_bits` is accepted, since it constrains the client's deflater
and `Inflater` copes with any legal window.

`client_no_context_takeover` and `server_no_context_takeover` are both honoured
when offered. An unknown parameter makes an offer unacceptable rather than
being ignored — ignoring it would mean agreeing to terms we did not implement.

## Implementation notes

`PerMessageDeflate` is one instance per connection and is not thread-safe,
which it does not need to be: outbound frames are serialised by
`AsyncChannel.send` (already `synchronized`), and inbound frames are decoded on
the single IO thread owning the connection's `WsAtta`.

`negotiate` returns null when the extension was not offered or was offered only
with parameters it cannot satisfy; null means "no extension" and the connection
behaves exactly as before.

## Testing

15 tests / 73 assertions covering negotiation (including offer lists,
unknown-parameter rejection, and both no-context-takeover directions),
round-trips with and without context takeover, RSV1 handling, fragmented
messages, control frames interleaved with compressed data, and the
post-inflation size bound. Driven through a real socket against a running
server rather than by calling the class directly.

Full suite green: 154 tests / 7330 assertions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ native-test-0.0.1-standalone.build_artifacts.txt
 target
 .lein-repl-history
 /wiki/.git
+
+# clj CLI classpath cache
+.cpcache/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,40 @@ This project uses [**Break Versioning**](https://www.taoensso.com/break-versioni
 
 ---
 
+# Unreleased
+
+## New: WebSocket `permessage-deflate` (RFC 7692)
+
+The server now accepts a client's `permessage-deflate` offer and compresses
+WebSocket messages with **context takeover**, so message N is compressed
+against messages 1..N-1. This is what makes it worth having for a stream of
+many small similar messages: compressing each message independently saves ~10%,
+while context takeover took a sample stream of 500 small JSON-ish messages from
+88 624 to 10 783 bytes (**-88%**).
+
+- Off unless the client offers it, so a connection that does not negotiate the
+  extension is byte-identical to before.
+- Bind `*websocket-compression?*` to false to refuse it. Worth doing when
+  payloads are already compressed (images, video), where deflate spends CPU to
+  make them slightly larger, or when per-connection memory matters more than
+  bandwidth -- context takeover keeps a deflate and an inflate window alive for
+  the life of each connection.
+- `*websocket-max-message-size*` bounds a message's size **after**
+  decompression. This is separate from `:max-ws`, which bounds the bytes
+  actually received and so says nothing about the size after inflation.
+
+A message whose compressed form is empty goes out as the single octet `0x00`
+per RFC 7692 7.2.3.6, not as zero bytes. Zero bytes round-trips fine on a fresh
+connection and desynchronises the stream mid-conversation, once there is
+compression history for it to corrupt.
+
+Not implemented: `server_max_window_bits`. `java.util.zip` does not expose
+zlib's windowBits, so the server cannot honour a smaller window and does not
+claim to -- the parameter is simply not echoed, which per RFC 7692 means a
+15-bit window. `client_max_window_bits` is accepted.
+
+---
+
 # `v2.9.0-beta4` (2026-07-31)
 
 - **Dependency**: [on Clojars](https://clojars.org/http-kit/versions/2.9.0-beta4)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,10 +29,13 @@ per RFC 7692 7.2.3.6, not as zero bytes. Zero bytes round-trips fine on a fresh
 connection and desynchronises the stream mid-conversation, once there is
 compression history for it to corrupt.
 
-Not implemented: `server_max_window_bits`. `java.util.zip` does not expose
-zlib's windowBits, so the server cannot honour a smaller window and does not
-claim to -- the parameter is simply not echoed, which per RFC 7692 means a
-15-bit window. `client_max_window_bits` is accepted.
+Not implemented, and therefore DECLINED rather than silently accepted:
+`server_max_window_bits`. `java.util.zip` does not expose
+zlib's windowBits, so the server cannot honour a smaller window, and per RFC
+7692 7.1.2.1 a server that does not support the parameter declines the offer --
+accepting while omitting it is not valid, and would leave a client that sized
+its inflate window at 1 KiB reading 32 KiB-distance references.
+`client_max_window_bits` is accepted.
 
 ---
 
--- a/src/java/org/httpkit/HttpUtils.java
+++ b/src/java/org/httpkit/HttpUtils.java
@@ -659,12 +659,22 @@ public class HttpUtils {
     }
 
     public static ByteBuffer WsEncode(byte opcode, byte[] data, int length) {
+        return WsEncode(opcode, data, length, false);
+    }
+
+    /**
+     * @param rsv1 set the RSV1 bit, which under RFC 7692 (permessage-deflate)
+     *             marks the message as compressed. Only ever legal on the first
+     *             frame of a data message, and never on a control frame.
+     */
+    public static ByteBuffer WsEncode(byte opcode, byte[] data, int length, boolean rsv1) {
         if ((opcode & 0x08) != 0 && length > 125) {
             throw new IllegalArgumentException(
                     "Websocket control frame payload exceeds 125 bytes");
         }
         byte b0 = 0;
         b0 |= 1 << 7; // FIN
+        if (rsv1) b0 |= 0x40;
         b0 |= opcode;
         int headerLength = length <= 125 ? 2 : length <= 0xFFFF ? 4 : 10;
         ByteBuffer buffer = ByteBuffer.allocate(length + headerLength);
@@ -685,6 +695,6 @@ public class HttpUtils {
     }
 
     public static ByteBuffer WsEncode(byte opcode, byte[] data) {
-        return WsEncode(opcode, data, data.length);
+        return WsEncode(opcode, data, data.length, false);
     }
 }
--- a/src/java/org/httpkit/server/AsyncChannel.java
+++ b/src/java/org/httpkit/server/AsyncChannel.java
@@ -47,6 +47,9 @@ public class AsyncChannel {
     // streaming
     private volatile boolean headerSent = false;
     private volatile boolean websocketUpgraded = false;
+    // Null unless permessage-deflate (RFC 7692) was negotiated. Guarded by the
+    // same monitor as send(), which is the only thing that touches it.
+    private PerMessageDeflate perMessageDeflate;
 
     // messages sent from a WebSocket client should be handled orderly by server
     // Changed from a Single Thread(IO event thread), no volatile needed
@@ -294,6 +297,14 @@ public class AsyncChannel {
             handler = closeHandler;
             ringHandler = closeRingHandler;
         }
+        // Deflater/Inflater hold native memory that is not reclaimed by GC
+        // promptly. A server holding many connections would otherwise leak it
+        // for as long as the objects stay reachable.
+        PerMessageDeflate pmd = getPerMessageDeflate();
+        if (pmd != null) {
+            setPerMessageDeflate(null);
+            pmd.end();
+        }
         if (handler != null) {
             handler.invoke(readable(status));
         }
@@ -341,6 +352,39 @@ public class AsyncChannel {
         return true;
     }
 
+    /**
+     * Install the negotiated permessage-deflate codec. Called once, from the
+     * handshake, before any frame is sent.
+     *
+     * <p>Also pushes it into the connection's decoder. The {@link WsAtta} that
+     * owns that decoder is created when the upgrade request is DECODED, which
+     * happens before the Ring handler and therefore before negotiation, so it
+     * cannot read the codec for itself -- both directions have to be wired from
+     * here, after there is something to wire.
+     */
+    public synchronized void setPerMessageDeflate(PerMessageDeflate pmd) {
+        this.perMessageDeflate = pmd;
+        Object atta = key.attachment();
+        if (atta instanceof WsAtta) {
+            ((WsAtta) atta).decoder.setPerMessageDeflate(pmd);
+        }
+    }
+
+    public synchronized PerMessageDeflate getPerMessageDeflate() {
+        return perMessageDeflate;
+    }
+
+    /** Frame a DATA payload, compressing it when the extension is active.
+     *  Control frames never come through here -- RFC 7692 6.1 forbids
+     *  compressing them. */
+    private ByteBuffer wsData(byte opcode, byte[] data, int length) {
+        if (perMessageDeflate != null) {
+            byte[] deflated = perMessageDeflate.compress(data, length);
+            return WsEncode(opcode, deflated, deflated.length, true);
+        }
+        return WsEncode(opcode, data, length);
+    }
+
     public synchronized boolean send(Object data, boolean close) throws IOException {
         if (closedRan.get()) {
             return false;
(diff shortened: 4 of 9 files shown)


whilo:
Pushed a fix for a defect I found after opening this, by running the server
against an independent client implementation (a Tyrus `ExtendedExtension`)
rather than against itself.

**RFC 7692 §7.2.3.6**: when the compressor produces no output, the payload must
be a single empty uncompressed DEFLATE block, `0x00` — not zero bytes. The
RFC's words: *"If the compression library being used doesn't generate any data
when its buffer is empty, an empty uncompressed DEFLATE block can be built and
used for this purpose as follows: 0x00"*.

I had it emitting zero bytes, with a comment confidently asserting that was
what the RFC prescribed. It isn't.

What makes it worth a regression test: zero bytes round-trips **fine** on a
fresh connection and fine as the first message, because there is no compression
history yet for it to corrupt. Send `["a" "" "b"]` over one connection with
context takeover and the stream desynchronises — `"b"` never arrives. A test
that sends an empty message in isolation passes, which is exactly what the
original suite did.

Added `empty-message-mid-stream`, which sends `["a" "" "b" "" "" "c"]` in
sequence and asserts both that no compressed payload is ever zero-length and
that each round-trips.

Suite: 155 tests / 7313 assertions.

Ramblurr (changes requested):
hi @whilo  I threw some tokens at this as a review and found three rfc correctness issues which seem legit to my manual reading of the rfc.

1. Quoted values are parsed as separate offers
   RFC 7692 §5.2 says, "The value may be quoted." Section 7 requires declining an offer containing
   "an extension parameter with an invalid value." The parser splits commas inside the quoted value
   and accepts a phantom offer.

2. Duplicate detection depends on locale
   RFC 7692 §7 requires declining when "the negotiation offer contains multiple extension parameters
   with the same name." Under the Turkish locale, locale-sensitive lowercasing lets differently cased
   duplicates through. I think the fix is to use `key.toLowerCase(Locale.ROOT)` instead.

 3. The server accepts a compressed message with no DEFLATE data

    RFC 7692 §7.2.3.6 says an empty compressed fragment must contain an empty DEFLATE block: "an empty
    uncompressed DEFLATE block can be built … as follows: 0x00."
    A compressed frame with a zero-byte payload does not contain that block. It is malformed. The
    current decoder accepts it and sends an empty message to the application. It should reject the
    frame and fail the WebSocket connection.


<details>
<summary>Here is a clojure test ns you can use as a jumping off point...<code>(ns org.httpkit.permessage-deflate-rfc-test..</code></summary>

```clojure
(ns org.httpkit.permessage-deflate-rfc-test
  (:require
   [clojure.test :refer [deftest is testing]])
  (:import
   [java.util Locale]
   [org.httpkit.server PerMessageDeflate WebSocketException]))

(deftest quoted-comma-does-not-create-another-offer
  (testing "an invalid quoted value must not become a second offer"
    ;; RFC 7692 section 5.2 says, "The value may be quoted." Section 7
    ;; requires the server to decline an offer containing a parameter with an
    ;; invalid value.
    ;;
    ;; This is one offer with an invalid quoted client_max_window_bits value.
    ;; The current parser splits at the quoted commas and mistakes the middle
    ;; text for a second, valid permessage-deflate offer.
    (let [offer (str "permessage-deflate; client_max_window_bits="
                     "\"10, permessage-deflate,11\"")
          ^PerMessageDeflate negotiated
          (PerMessageDeflate/negotiate offer (* 1024 1024))]
      (try
        (is (nil? negotiated)
            "RFC 7692 section 7 requires the malformed offer to be declined")
        (finally
          (when negotiated
            (.end negotiated)))))))

(deftest duplicate-parameter-names-are-declined-in-every-locale
  (testing "duplicate names remain duplicates under the Turkish locale"
    ;; RFC 7692 section 7 requires the server to decline an offer that
    ;; "contains multiple extension parameters with the same name."
    ;;
    ;; String.toLowerCase uses the JVM default locale. Under Turkish, the two
    ;; spellings below normalize differently even though equalsIgnoreCase
    ;; correctly recognizes both names.
    (let [original-locale (Locale/getDefault)]
      (try
        (Locale/setDefault (Locale/forLanguageTag "tr-TR"))
        (let [offer (str "permessage-deflate; CLIENT_MAX_WINDOW_BITS=10; "
                         "client_max_window_bits=11")
              ^PerMessageDeflate negotiated
              (PerMessageDeflate/negotiate offer (* 1024 1024))]
          (try
            (is (nil? negotiated)
                "RFC 7692 section 7 requires duplicate parameters to be declined")
            (finally
              (when negotiated
                (.end negotiated)))))
        (finally
          (Locale/setDefault original-locale))))))

(deftest zero-byte-compressed-payload-is-rejected
  (testing "an empty compressed payload contains the required 0x00 octet"
    ;; RFC 7692 section 7.2.1 says that after compression, "the last octet of
    ;; the compressed data contains ... the DEFLATE header bits with the BTYPE
    ;; bits set to 00." Section 7.2.3.6 gives that empty block as 0x00.
    ;;
    ;; The current decoder accepts zero bytes and delivers an empty message.
    (let [^PerMessageDeflate negotiated
          (PerMessageDeflate/negotiate "permessage-deflate" (* 1024 1024))]
      (try
        (is (= 1002
               (try
                 (.decompress negotiated (byte-array 0))
                 :accepted
                 (catch WebSocketException error
                   (.getCloseStatus error))))
            "a zero-byte compressed payload must fail with protocol status 1002")
        (finally
          (.end negotiated))))))
```

</details>

whilo:
Hey @Ramblurr, thanks for helpfing with this! I worked through your feedback, studied RFC 7230, ws, Jetty, ... in multiple passes and landed on this result with the following reply from Claude. I also ran Autobahn's test suite locally, it might be reasonable to add to CI/CD, not sure.

Claude:

@Ramblurr thank you — all three were real, and chasing them turned up four more
that were worse. Pushed as 8 commits.

## Your three

1. **Quoted values parsed as separate offers.** Confirmed. The header was split
   on `,` and `;` before any quote handling, so `client_max_window_bits="10,
   permessage-deflate,11"` — one offer with one invalid value — was cut into
   three, and the middle fragment looked like a second, bare, acceptable offer.
   Replaced with a quoted-string-aware splitter that also handles RFC 7230
   quoted-pairs and declines a header with unbalanced quotes. A legitimately
   quoted value (`client_max_window_bits="12"`, permitted by §5.2) now works
   too, which the old parser only got right by accident.

2. **Locale-dependent duplicate detection.** Confirmed, and your suggested fix
   was the right one: `toLowerCase(Locale.ROOT)`. `equalsIgnoreCase` elsewhere
   was already locale-independent; only that fold wasn't.

3. **Zero-byte compressed payload.** Confirmed, and worse than "delivers an
   empty message": appending the required tail to nothing yields `00 00 FF FF`,
   whose leading `0x00` the inflater reads as a stored-block header and whose
   next octets it reads as `LEN = 0xFF00`. It then sits mid-block, so the
   *next* message dies with `invalid stored block lengths` — one message away
   from the cause. Now rejected with 1002 at the frame that caused it.

   One refinement on the normative basis, since it shaped how far we took it:
   §7.2.3.6 sits inside §7.2.3 *Examples* and is written for a sender, so
   there's no receiver-side MUST to quote. The hook is §7.2.1 — the payload is
   DEFLATE blocks minus the 4-octet tail, and steps 2–3 guarantee it ends on a
   block header, so a conformant payload always has at least one octet. That
   framing is what let us draw a consistent line elsewhere: reject what
   desynchronises the stream, tolerate what is merely sloppy.

Your test namespace is in, expanded.

## Four more, found by reviewing against the RFC and ws/Netty/Jetty/Tyrus

- **A `BFINAL=1` block silently emptied the rest of the connection.** §7.2.3.4
  explicitly permits flushing that way and gives the wire bytes. Once
  `Inflater` reports `finished()` it returns 0 forever, so every later message
  was delivered to the application as an empty string with no error at either
  end — an 8-byte frame kills the connection. §7.2.1's "the next DEFLATE block
  follows" case was also truncating within a message. For what it's worth,
  **Jetty still has this**; ws recreates its inflater and Netty fails loudly.
- **Server-initiated close broke the closing handshake.** The codec was
  released when we *sent* CLOSE, but the socket stays open awaiting the peer's
  (RFC 6455 §5.5.1), so a compressed frame already in flight became
  "unsupported extension data" — a clean 1000 became 1000, then 1002, then a
  reset, only on compressed connections.
- **A whole-server deadlock**, introduced by that close fix: releasing the
  codec from `closeKey` took the `AsyncChannel` monitor while `doWrite` held
  the `ServerAtta` monitor, inverting the order `send` uses. It hung the single
  selector thread — every connection, compressed or not, since the release path
  runs regardless. Reproduced with `ThreadMXBean`; fixed lock-free, 401 rounds
  clean.
- **A DoS in the fix for the above.** DEFLATE stream restarts cost a native
  `Inflater.reset()` that produces no output, so the post-inflation size guard
  couldn't see them: a 4 MiB frame of two-octet streams cost 2–3 orders of
  magnitude more IO-thread time than a legitimate 4 MiB message. Now bounded by
  two ceilings, failing with 1008 (policy, not conformance — §7.2.1 sets no
  limit on blocks per message, and messages with thousands of data-bearing
  streams are still accepted).

Being straight about it: four of the seven were self-inflicted by earlier
fixes in this branch, and several code comments asserted things that later
measurement disproved. They're corrected, and the risky paths now have
regression tests that fail without their fix.

## Also in this branch

- `server_max_window_bits=15` is now accepted and echoed — satisfiable exactly,
  since our deflater's window *is* 15. That alone moves 36 Autobahn cases from
  UNIMPLEMENTED to OK. Smaller values are still declined.
- `*websocket-compression-threshold*`, defaulted to **0** rather than copying
  ws's 1024: with context takeover, 16-byte JSON-ish messages measure 0.25x and
  64-byte English text 0.07x, so a 1024-byte threshold would discard exactly
  the win the extension exists for.
- `*websocket-max-message-size*` derives from `:max-ws`, so lowering that knob
  doesn't silently stop bounding memory for compressed clients.
- One unrelated fix in its own commit, easy to drop or cherry-pick: unassigned
  close codes in 1016–2999 were echoed back instead of failing with 1002
  (Autobahn 7.9.6–7.9.9, failing on master). **Breaking**: `close` with such a
  status now throws.

## Verification

Autobahn Testsuite against this branch: **301 base cases and 216 compression
cases, 0 failures** (base had 4 before the close-code fix). `lein test` 176
tests / 0 failures.

Two pre-existing issues surfaced along the way that I deliberately did *not*
touch here, both verified present on master: `closeKey → clientClose` can run
`onClose` inline under the `ServerAtta` monitor, so a user close handler
calling `send`/`close` can deadlock; and the selector loop doesn't close keys
when it exits via `Error`. Happy to open issues for both.

ptaoussanis:
@whilo Hi Christian, thanks for this! A few thoughts from me:

1. Looks like `*websocket-compression?*` defaults to enabled by default, right? Could you motivate that? My inclination would be an explicit `run-server` opt defaulting to false (at the very least until the feature is widely tested and we're confident it wouldn't cause any regressions).

2. Inflation runs on http-kit's single selector thread, is that right? Any concern there?

3. Please don't add detailed docs to http-kit's CHANGELOG. Better find an appropriate spot in the wiki and/or relevant commit message, then the CHANGELOG can link there.

4. Looks like there's a fair bit of stuff committed related to review history, benchmarking anecdotes, etc. That stuff should please be kept to PR discussion and out of the committed work.

Also ran this by Codex, which highlighted a few potential technical concerns to double check:

> A. It mishandles valid repeated extension headers.
     HTTP-kit combines duplicate request headers with "\n", but the new parser only splits offers on commas. RFC 6455
     explicitly permits:

     Sec-WebSocket-Extensions: unsatisfiable-offer
     Sec-WebSocket-Extensions: permessage-deflate

     and defines it as equivalent to a comma-separated value. RFC 6455 §9.1
     (https://datatracker.ietf.org/doc/html/rfc6455#section-9.1)

     At current HEAD, that valid fallback is declined. I reproduced it directly. This should have a regression test and likely
     just treat http-kit’s internal newline separator like a comma.


> B. The scope contains an unrelated breaking change.
     Tightening WebSocket close-code validation is defensible, but it is unrelated and explicitly breaking. It should be its
     own PR/commit, not bundled with compression.


Thanks!

whilo:
Thanks @ptaoussanis — all four points taken. Force-pushed as a single commit
(`e074cac`); the branch is now one clean commit against master.

**1. Off by default, and as a `run-server` opt.** Agreed, and I went a little
further: all three knobs are now `run-server` options rather than dynamic vars.

```clojure
:websocket-compression?           ; default false
:websocket-max-message-size       ; default :max-ws
:websocket-compression-threshold  ; default 0
```

Besides your reasoning, the dynamic vars were read on the *handler* thread
during the handshake, so a user's `binding` never reached them — our own tests
had to resort to `with-redefs`. A per-server option is both what you asked for
and simply more correct. All three were new in this PR, so nothing to migrate.
There's a test that a default server declines a perfectly good offer.

**2. Inflation on the selector thread.** Yes, that's right, and worth being
explicit about. It's bounded — by `:websocket-max-message-size` after
inflation, and by ceilings on DEFLATE streams per message that the size bound
can't see (a stream restart produces no output) — but a large message is still
real work on the shared IO thread: roughly 15-25 ms for a 4 MiB message here,
microseconds for typical traffic.

Moving it into the ordered per-connection handler task is doable, but it drags
UTF-8 validation and protocol-error detection off the selector thread with it,
which felt like the wrong thing to bundle into this PR. With the feature now
opt-in the exposure is a deliberate choice by the operator, and it's documented
in the wiki. Happy to do it as a follow-up if you'd like it before this is
enabled by default.

**3. CHANGELOG.** Cut from ~90 lines to 14, pointing at a new *WebSocket
compression* section in `wiki/3-Server.md` (options table, context-takeover
rationale with the measurements, costs, and what isn't supported). Followed the
`:legacy-unsafe-remote-addr?` precedent of a short entry plus a wiki link.

**4. Review history and anecdotes out of the committed work.** Fair, and
thoroughly deserved — the commit messages were essentially review transcripts
and the code comments carried a lot of "an earlier version was wrong about X".
Squashed to one commit, and the comments now explain only why the code is the
way it is. `PerMessageDeflate` lost ~80 lines of narrative that way. Also
dropped a stray `.gitignore` entry that had nothing to do with this.

### On Codex's two points

**B. Close codes → own PR.** Agreed, done: #620. That also means this PR no
longer changes base-suite behaviour at all — the four Autobahn 7.9.x failures
it fixed are back to failing here exactly as they do on master, and #620 fixes
them.

**A. Repeated `Sec-WebSocket-Extensions` headers.** I couldn't reproduce this
at HEAD, and I think the analysis missed your own #615. Against the branch:

```
Sec-WebSocket-Extensions: unsatisfiable-offer
Sec-WebSocket-Extensions: permessage-deflate
   -> Sec-Websocket-Extensions: permessage-deflate
```

`ringHeaders` normalises the internal `\n` join to `,` before the Ring map,
which is what negotiation reads, so the parser sees a normal comma-separated
list. Codex looks to have reasoned from the raw `HeaderMap` (or from a tree
without #615).

That said, the fix it suggested costs nothing and makes the parser correct
regardless of caller, so I took it anyway: the splitter now treats `\n` and
`\r` as offer separators alongside `,`, with a regression test covering the
newline-joined form, the "skip the unsatisfiable one, take the next" case, and
the comma-joined form the Ring map actually delivers.

### Verification after the rewrite

- Autobahn: **216 compression cases, 0 failures**; base suite unchanged from
  master (the 4 close-code cases fail here and on master, fixed in #620).
- `lein test`: 175 tests, 0 failures.

Unrelated, in case it's news: `org.httpkit.server-test/test-server-status`
fails on pristine master when run in isolation (`:stopping` vs `:stopped`).
Nothing to do with either PR.

